PR #6, opened 2015-03-18
feat(metaconfig): for new config sources, add groupable options and refactor mutual exclusion

samstav:
meta config options allow an entire configuration to be
specified by a single source. the only existing native
capability was the use of ini files, but there was no
way to specify which ini file, much less specify it the
same way any other Option could be specified (cli, env, keyring).

this connects a metaconfig instance that can be used to provision
an entire (or part of) a Config. the first meta option to be
implemented was `--ini` since Config natively supports parsing ini
files.

See the MetaConfig docstring for more details.

#### todo

- [x] tests

### DIFF
--- a/simpl/config.py
+++ b/simpl/config.py
@@ -219,7 +219,7 @@ class Option(object):
         updater = {k: v for k, v in copy.copy(self.__dict__).items()
                    if k not in ('args', 'kwargs')}
         newone.__dict__.update(updater)
-        assert not newone.kwargs is self.kwargs
+        assert newone.kwargs is not self.kwargs
         return newone
 
     def __repr__(self):
@@ -335,7 +335,8 @@ class Config(collections.MutableMapping):
 
     """Parses configuration sources."""
 
-    def __init__(self, options=None, ini_paths=None, argv=None, **parser_kwargs):
+    def __init__(self, options=None, ini_paths=None, argv=None,
+                 **parser_kwargs):
         """Initialize with list of options.
 
         :param ini_paths: optional paths to ini files to look up values from
@@ -593,7 +594,8 @@ class Config(collections.MutableMapping):
                 if option.dest not in results or results[option.dest] is None:
                     if getattr(option, '_mutexgroup', None):
                         raise_for_group.setdefault(option._mutexgroup, [])
-                        raise_for_group[option._mutexgroup].append(option._action)
+                        raise_for_group[option._mutexgroup].append(
+                            option._action)
                     else:
                         raise SystemExit("'%s' is required. See --help "
                                          "for more info." % option.name)
@@ -693,7 +695,8 @@ class MetaConfig(Config):
     """
 
     option_group = 'initialization (metaconfig) arguments'
-    option_description = 'evaluated first and can be used to source an entire config'
+    option_description = ('evaluated first and can be used to '
+                          'source an entire config')
 
     options = [
         Option('--ini', metavar='PATH',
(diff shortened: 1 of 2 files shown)


ziadsawalha:
The @argument at the end I think supported supplying an ini file per [the docs](https://github.com/checkmate/simpl/blob/master/simpl/config.py#L127). Was that not sufficient?

simpl/config.py
```diff
@@ -360,7 +395,8 @@ def parse_cli(self, argv=None, permissive=False):
             kwargs['default'] = argparse.SUPPRESS
             temp = Option(*option.args, **kwargs)
             options.append(temp)
-        parser = self.build_parser(options, permissive=permissive)
+        parser = self.build_parser(
+            options, permissive=permissive)
```
ziadsawalha: This change is not necessary. The line fits within PEP8 limits.

simpl/config.py
```diff
@@ -298,6 +300,39 @@ def prog(self):
...
+        """Default ini file name."""
+        return '%s.ini' % self.prog
+
+    def _metaconfigure(self):
+        """Initialize metaconfig for provisioning self."""
+        metaconfig = self._get_metaconfig_class()
+        if not metaconfig:
+            return
+        if self.__class__ is metaconfig:
+            # dont get too meta
```
ziadsawalha: minor spelling - missing `'`

simpl/config.py
```diff
@@ -298,6 +300,39 @@ def prog(self):
...
+        metaparser = self._metaconf.build_parser(
+            options=self._metaconf._options, permissive=False, **override)
+        for grp in metaparser._action_groups:
+            grp.title = 'initialization (metaconfig) arguments'
+        self._parser_kwargs.setdefault('parents', [])
+        self._parser_kwargs['parents'].append(metaparser)
+        self._metaconf.parse()
+        self._metaconf.provision(self)
+
+    def _get_metaconfig_class(self):
```
ziadsawalha: This could be static or a classmethod that returns `cls`.

simpl/config.py
```diff
@@ -408,16 +444,35 @@ def parse_ini(self, paths=None):
...
+                    print('Error parsing ini file: %r -- Continuing.'
+                          % err, file=sys.stderr)
+                except ConfigParser.NoOptionError as err:
+                    print('Error parsing ini file: %r -- Continuing.'
+                          % err, file=sys.stderr)
+            if not value:
+                try:
+                    value = config.get(namespace, option.name)
+                    results[option.dest] = option.type(value)
+                except ConfigParser.NoSectionError as err:
```
ziadsawalha: `except (ConfigParser.NoSectionError, ConfigParser.NoOptionError):` or if there is a common superclass? `err` is not used.

simpl/config.py
```diff
@@ -408,16 +444,35 @@ def parse_ini(self, paths=None):
...
         for option in self._options:
             ini_section = option.kwargs.get('ini_section')
+            value = None
             if ini_section:
                 try:
                     value = config.get(ini_section, option.name)
                     results[option.dest] = option.type(value)
-                except ConfigParser.NoSectionError:
+                except ConfigParser.NoSectionError as err:
+                    print('Error parsing ini file: %r -- Continuing.'
```
ziadsawalha: I don't know that this class should be assuming it can own outputting to stdout. It should probably LOG and let the app or deployer decide how that output should be routed. Especially if I have `--quiet` as an argument.

simpl/config.py
samstav: There is a `ConfigParser.Error` but it is the parent for too many others.

simpl/config.py
```diff
@@ -530,3 +719,25 @@ def comma_separated_pairs(value):
 def parse_key_format(value):
     """Handle string formats of key files."""
     return value.strip("'").replace('\\n', '\n')
+
+
+if __name__ == '__main__':
+
+    opts = [
+        Option('--foo'),
```
ziadsawalha: Is this a local test that made it into the PR?

Gifflen: From what I gathered by fe1002e it is example functionality.

Gifflen:
I really like the change to the mutual exclusion to allow groups to exist separately.  Cleaning of the CLI help to be more user friendly is big win.

PR #13, opened 2015-05-22
threadlocal util + tests

samstav:
Closes https://github.com/checkmate/simpl/issues/14


tests/test_threadlocal.py
```diff
@@ -0,0 +1,87 @@
...
+        for _ in xrange(num_threads):
+            t = threading.Thread(target=get_tld, args=(tcon,))
+            t.start()
+            joins.append(t)
+        # and one to grow on
+        threadlocal.get_context()[key] = self.get_some_text()
+        tcon.put(threadlocal.get_context()._get_local_dict())
+
+        for thrd in joins:
+            t.join()
```
ziadsawalha: Should this be `thrd.join()`?

tests/test_threadlocal.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class TestThreadLocal(unittest.TestCase):
+
+    def setUp(self):
+        # clear the threadlocal dict
+        threadlocal.get_context().clear()
+
+    def get_some_text(self, length=16):
+
```
ziadsawalha: Methods & functions should have docstrings (PEP-8) and there should be no blank lines before or after docstrings for methods/functions (PEP-257). So even if we omit the docstring in a test module, I think the blank lines at that beginning of the method should be removed.

simpl/threadlocal.py
```diff
@@ -0,0 +1,52 @@
...
+    """A dict whose data is local to the thread."""
+
+    def __init__(self, varname, *args, **kwargs):
+        self.varname = varname
+        self.args = args
+        self.kwargs = kwargs
+
+    def __repr__(self):
+
+        under = repr(self._get_local_dict())
```
ziadsawalha: Curiosity killed the cat, I know, but I can't help but ask... why is the variable called `under`?

tests/test_threadlocal.py
```diff
@@ -0,0 +1,87 @@
...
+        key = 'value'
+        num_threads = 5
+
+        def get_tld(container):
+            current = threadlocal.get_context()
+            current[key] = self.get_some_text()
+            container.put(current._get_local_dict())
+
+        tcon = Queue.Queue()
+        joins = []
```
ziadsawalha: IMO, this code would be a bit easier to understand when first read if these variables were `local_dicts` and `threads`.

```python
local_dicts = Queue.queue()
threads = []
```

tests/test_threadlocal.py
```diff
@@ -0,0 +1,87 @@
...
+            values.add(tld[key])
+
+        # assert 6 dicts were created with 6 different values
+        self.assertEqual(len(values), num_threads+1)
+
+    def test_non_default_namespace(self):
+
+        tld = threadlocal.get_context()
+        another = threadlocal.ThreadLocalDict('custom_namespace')
+        self.assertIsNot(tld._get_local_dict(), another._get_local_dict())
```
ziadsawalha: This is a negative test. Should there also be a positive one that checks that the dicts are the same?

simpl/threadlocal.py
```diff
@@ -0,0 +1,52 @@
...
+THREAD_STORE = threading.local()
+DEFAULT_NAMESPACE = 'call_context'
+
+
+class ThreadLocalDict(collections.MutableMapping):
+
+    """A dict whose data is local to the thread."""
+
+    def __init__(self, varname, *args, **kwargs):
+        self.varname = varname
```
ziadsawalha: Would `namespace` instead of `varname` describe this variable better?

ziadsawalha: Statically putting the class name in the repr string would display an incorrect class name when subclassed. `return '<%s %s>' % (self.__class__.__name__, under)` would work even if subclassed.

simpl/threadlocal.py
```diff
@@ -0,0 +1,52 @@
...
+        self.args = args
+        self.kwargs = kwargs
+
+    def __repr__(self):
+
+        under = repr(self._get_local_dict())
+        return '<ThreadLocalDict %s>' % under
```
ziadsawalha: Statically putting the class name in the repr string would display an incorrect class name when subclassed. `return '<%s %s>' % (self.__class__.__name__, under)` would work even if subclassed.

simpl/threadlocal.py
```diff
@@ -0,0 +1,52 @@
+"""Threadlocal."""
```
ziadsawalha: config.py has a long description with examples of how to use it. That woud be a good pattern to continue in simpl; to have each module documented in the __doc__.

simpl/threadlocal.py
```diff
@@ -0,0 +1,52 @@
...
+        under = repr(self._get_local_dict())
+        return '<ThreadLocalDict %s>' % under
+
+    def _get_local_dict(self):
+        """Retrieve (or initialize) the thread-local data to use."""
+        local_var = getattr(THREAD_STORE, self.varname, None)
+        if not local_var:
+            local_var = dict(*self.args, **self.kwargs)
+            setattr(THREAD_STORE, self.varname, local_var)
+        return local_var
```
ziadsawalha: If we use the presence of the attr as the test, then the thread local dict only gets initialized once (and can be falsy):
```python
    def _get_local_dict(self):
        """Retrieve (or initialize) the thread-local data to use."""
        try:
            return getattr(THREAD_STORE, self.varname)
        except AttributeError:
            local_var = dict(*self.args, **self.kwargs)
            setattr(THREAD_STORE, self.varname, local_var)
            return local_var
```

simpl/threadlocal.py
```diff
@@ -0,0 +1,52 @@
...
+    def __getitem__(self, key):
+        return self._get_local_dict()[key]
+
+    def __setitem__(self, key, value):
+        self._get_local_dict()[key] = value
+
+    def __delitem__(self, key):
+        self._get_local_dict().__delitem__(key)
+
+
```
ziadsawalha: I'm thinking about how to simplify the usage of this function from outside simpl. I wonder if everything below this line (and the definition of `DEFAULT_NAMESPACE` above) should live outside this module.

```python
import uuid

import simpl.context

CONTEXT = simple.context.default()

def some_thing(context):
    """Do something with context.""""
    context['transactionId'] = uuid.uuid4().hex

some_thing(CONTEXT)
print context.default()
#>>> <ThreadLocalContext {'transactionId': 'badabadabadabadabada'}>

# Custom context
print simple.context.new('custom_context').setdefault('foo', 1)
#>>> <ThreadLocalContext {'foo': 1}>
```

Not sure yet.... thinking out loud :-)

tests/test_threadlocal.py
samstav: Yes. Hahaha. Funny that the tests work anyway... Maybe that means I don't need thus loop. Thanks!

simpl/threadlocal.py
samstav: Good point. I'll add some here

tests/test_threadlocal.py
samstav: Agreed. Had a weird time writing this test...

simpl/threadlocal.py
samstav: It is the dict underneath the object backing all methods? idk.

simpl/threadlocal.py
samstav: Yes

simpl/threadlocal.py
samstav: +1 I like

simpl/threadlocal.py
samstav: I like the `default()` and `new()` interface. Clean and intuitive. But if the `default()` method is to function and return a threalocal dict, that seems to imply that `DEFAULT_NAMESPACE` is defined in the module. 

```
import simpl.context

print simpl.context.default()
# <ThreadLocalDict {}>
```

We know that the actual dictionary there is an attribute of the threadlocal object, but what would the attribute name be?

samstav:
Some notes:

The specific instance of `ThreadLocalDict` that you might have doesn't really matter. If you bring that instance into a different thread it will immediately start referencing a different underlying threadlocal object. 

```python


import threading

from simpl import threadlocal

TLD = threadlocal.ThreadLocalDict('GreatNamespace')
print 'TLD: %s' % TLD

def mutate(tld):
    tld['value'] = 'foo'
    print 'tld: %s' % tld
    assert tld is TLD
    assert tld

t = threading.Thread(target=mutate, args=(TLD,))
t.start()
t.join()

print 'TLD: %s' % TLD
# this works because TLD now references its original empty dict from the main thread
assert not TLD
```

prints:
```
TLD: <ThreadLocalDict {}>
tld: <ThreadLocalDict {'value': 'foo'}>
TLD: <ThreadLocalDict {}>
```

samstav:
I was just writing tests for checking equivalence, i.e. `==`. As of right now, if the underlying dictionaries are equivalent, the `ThreadLocalDict` instances return `True` for `==`. This hold true even if the namespaces are different. This sounds right to me and I've written tests to ensure that behavior. 

نعم؟

ziadsawalha:
!كول

PR #16, opened 2015-05-29
WIP READY FOR REVIEW: Conditional requirements utilities

samstav:
This should be reviewed and merged before any other PRs.

##### todo

- [ ] tests

### DIFF
--- a/simpl/utils/importing.py
+++ b/simpl/utils/importing.py
@@ -1,0 +1,85 @@
+"""Utilities for importing.
+
+The function import_me_maybe() uses the signature
+of the importlib.import_module() function.
+
+Usage:
+
+    from simpl.utils import import_me_maybe
+
+    eventlet = import_me_maybe('eventlet')
+    # this issues a warning (only once) if
+    # eventlet is not found
+
+    # to prevent warnings from ever occurring
+    from simpl.utils import disable_warnings_for
+
+    # this disables the corresponding warnings globally
+    disable_warnings_for('eventlet')
+
+    # this could be in another module
+
+    eventlet = import_me_maybe('eventlet')
+    # produces no warning now
+"""
+import importlib
+import inspect
+import os
+import warnings
+
+from simpl.exceptions import DependencyRequiredWarning
+
+
+def import_me_maybe(name, package=None):
+    """Try to import a module by name and return it.
+
+    If the module is not found, this function returns None.
+
+    Emits a DependencyRequiredWarning if the import fails along
+    with information about the caller that needed it.
+    Upon success, the specified module will be inserted into
+    sys.modules and returned.
+
+    The name argument specifies what module to import in absolute
+    or relative terms (e.g. either pkg.mod or ..mod). If the name
+    is specified in relative terms, then the package argument must
+    be specified to the package which is to act as the anchor for
+    resolving the package name
+    (e.g. import_module('..mod', 'pkg.subpkg') will import pkg.mod).
+    """
+    # lookup caller
+    importer = inspect.stack()[1]
+    importer_frame = importer[0]
+    importer_module_name = inspect.getmodulename(importer[1])
+    # so this behaves decently in an interpreter
+    if importer_module_name:
+        if hasattr(inspect, 'getabsfile'):
+            importer_path = inspect.getabsfile(importer_frame)
+        else:
+            importer_path = os.path.abspath(inspect.getfile(importer_frame))
+    else:
+        importer_path = importer[1]
+    fq_path = importer_path.split(os.path.sep)
+    if 'simpl' in fq_path:
+        # find the rightmost instance of 'simpl' in path
+        begin = (len(fq_path) - 1) - fq_path[::-1].index('simpl')
+        import_string = '.'.join(fq_path[begin:-1] + [importer_module_name])
+    else:
+        import_string = importer_module_name or importer[1]
+    try:
+        mod = importlib.import_module(name, package=package)
+    except ImportError as err:
+        mod = None
+        imp_err = '%s: %s' % (type(err).__name__, str(err))
+        warn_msg = DependencyRequiredWarning.format_msg(
+            import_string=import_string, requirement=name, from_exc=imp_err)
+        warnings.warn(warn_msg, DependencyRequiredWarning, stacklevel=2)
+    # it would be fun to inject the dependency here!
+    # importer_globals = importer_frame.f_globals
+    # importer_globals[name] = mod
+    return mod
+
+
+def disable_warnings_for(name):
+    """Disable DependencyRequiredWarning for this requirement."""
+    return DependencyRequiredWarning.filter(requirement=name)
--- a/simpl/exceptions.py
+++ b/simpl/exceptions.py
@@ -21,6 +21,8 @@ calling the disable() classmethod.
 import warnings
 
 __all__ = [
+    'SimplWarning',
+    'DependencyRequiredWarning',
     'GitWarning',
     'SimplException',
     'SimplGitError',
@@ -30,14 +32,28 @@ __all__ = [
 ]
 
 
-class GitWarning(RuntimeWarning):
+class SimplWarning(Warning):
 
-    """The local git program is missing or may be incompatible."""
+    """Base class for all Simpl warnings."""
 
     @classmethod
     def disable(cls):
-        """Disable warnings of this type."""
+        """Disable all warnings of this type."""
         return warnings.simplefilter('ignore', cls)
+
+
+class DependencyRequiredWarning(SimplWarning, ImportWarning):
+
+    """The simpl module requires a missing dependency."""
+
+# ImportWarning is disabled by default, make this warn
+warnings.simplefilter('default', DependencyRequiredWarning)
+
+
+class GitWarning(SimplWarning, RuntimeWarning):
+
+    """The local git program is missing or may be incompatible."""
+
 
 # shown until proven ignored :)
 warnings.simplefilter('always', GitWarning)
--- a/simpl/exceptions.py
+++ b/simpl/exceptions.py
@@ -18,6 +18,7 @@ Warnings can be imported and subsequently disabled by
 calling the disable() classmethod.
 """
 
+import functools
 import warnings
 
 __all__ = [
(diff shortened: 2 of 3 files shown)


simpl/utils/importing.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+import importlib
+import inspect
+import os
+import warnings
+
+from simpl.exceptions import DependencyRequiredWarning
+
+
+def import_me_maybe(name, package=None):
```
samstav: I've been thinking about adding a `reason` keyword argument here which tells the user why the module is needed. In some cases, we might emit this warning when the requirement is not strictly required. E.g. `eventlet` for the upcoming `server.py`, which I was thinking would run or be importable whether or not eventlet was installed. That might be useful.... `import_me_maybe('eventlet', reason='Required for non-blocking eventlet server')` which would append (or replace?) the generic message built-in and created by the `DependencyRequiredWarning` class.

ziadsawalha: +1 to informative warnings.
If `import_me_maybe` only warns the first time, how do we handle the warning if the second attempt has a different reason?

ziadsawalha:
@samstav - check out http://xion.org.pl/2012/05/06/hacking-python-imports/

ziadsawalha:
I think I'm also looking for an example of the intended use of this PR. Two concerns to consider:
1 - usage and style checkers.
2 - usage and behavior

```python
from simpl.utils.importing import import_me_maybe

# WRT to usage and style
eventlet = import_me_maybe('eventlet')  # Style checkers will complain about variable name.

# Alternatively, an import hook might let us do this
from maybe import eventlet  # where `maybe` is handled by a hook
# But explicit is better than implicit....


# WRT to behavior...
eventlet.monkey_patch()  # How do we want this to behave? What have we gained by import_me_maybe if it returns None and this fails?
```

simpl/utils/importing.py
samstav: The 'default' (once) warnings filter will warn once based on the origin of the warning. So I would assume that a different reason means different line of code where `import_me_maybe` is called, so you would get both warnings.

simpl/utils/importing.py
```diff
@@ -0,0 +1,85 @@
...
+        mod = importlib.import_module(name, package=package)
+    except ImportError as err:
+        mod = None
+        imp_err = '%s: %s' % (type(err).__name__, str(err))
+        warn_msg = DependencyRequiredWarning.format_msg(
+            import_string=import_string, requirement=name, from_exc=imp_err)
+        warnings.warn(warn_msg, DependencyRequiredWarning, stacklevel=2)
+    # it would be fun to inject the dependency here!
+    # importer_globals = importer_frame.f_globals
+    # importer_globals[name] = mod
```
samstav: @ziadsawalha WRT to complaints for variable names, I was using these lines to add the imported module into the caller's namespace. This is effectively what the `import` statement does too. 

```python
importer = inspect.stack()[1]
importer_frame = importer[0]
...
importer_globals = importer_frame.f_globals
importer_globals[name] = mod
```

I commented this out because this of course would not work if your imported package/module/object contains a `.`. For example

```
import_me_maybe('os.path')
```
I couldn't think of a relatively easy solution for this so I left that magic feature out entirely.

samstav:
Re: https://github.com/checkmate/simpl/pull/16#issuecomment-106789872

> WRT to usage and style
I'm not too worried about style things. Any style issues incurred are probably better that what I was trying to avoid, which was:

```python
try:


# Alternatively, an import hook might let us do this
from maybe import eventlet  # where `maybe` is handled by a hook
# But explicit is better than implicit....


# WRT to behavior...

**Comment by ziadsawalha** [comment]
But if we are not going to use a module, should we instead avoid importing it? And if we do import it, then we might want to know about missing dependencies before we actually try to use it.

Silent fails are OK if we have a fallback plan for the module not importing. Like:
```python
try:
    from eventlet import sleep
except ImportError:
    from time import sleep  # fallback
```

I guess that's why I'm asking where we would use this in simpl to be able to comment in an informed way on this PR.

samstav:
@ziadsawalha Are you asking what the purpose of this PR is? I am not sure if that is what you are getting at. It is primarily to avoid writing code like this:

```python
try:
    import eventlet
except ImportError:
    eventlet = None

if not eventlet:
    from simpl.exceptions import DependencyRequiredError
    warnings.warn("This module blah.blah and a lot of other hard-coded strings "
                 "requires eventlet for blah blah blah", DependencyRequiredError)

# and repeat these 10 lines of code for every conditional import, ewww....
# the purpose of import_me_maybe() was to 1) replace all this code and 2) do it even better
```

When this does the equivalent:

```
import_me_maybe('eventlet')
```
> But if we are not going to use a module, should we instead avoid importing it?

Yes of course. I must be misunderstanding the question.

Since not all "requirements" have a fallback, some missing dependencies will cause features to disappear. Much like the way keyring support disappears from `simpl/config` if the keyring import fails. (currently silently, that should change). 

And sure, if there's a good fallback plan, like your example with `sleep` or `ujson` vs `json`, `import_me_maybe` wouldn't need to be used there.

> we might want to know about missing dependencies before we actually try to use it

I think import time is the best time? It means you always get warned before you call anything in the module that will break (again, simpl module imports should never break, only warn) and you also avoid the warnings as long as you never import that specific module.

ziadsawalha:
So we would use this PR's features to replace [the following](https://github.com/checkmate/simpl/blob/master/simpl/config.py#L164-L167) code in config.py:
```python
try:
    import keyring
except ImportError:
    keyring = None  # pylint: disable=C0103
```
with one of these two:
- with magic
```python
from simpl.utils.importing import import_me_maybe
import_me_maybe('keyring')
```
- without magic
```python
from simpl.utils.importing import import_me_maybe
keyring = import_me_maybe('keyring')  # pylint: disable=C0103
```

samstav:
Yes I think so. Unless we look into the import hooks stuff. What do you think ?

ziadsawalha:
In the comparison above, I don't think we need the `if eventlet:`. Instead, we could write:
```python
import warnings

try:
    import eventlet
except ImportError:
    eventlet = None  # pylint: disable=C0103
    warnings.warn("Module blah.blah requires eventlet for optimal "
                  "performance. Eventlet is not available so a slower"
                  "algorithm will be used.")
```

and using `import_me_maybe()` to achieve a similar outcome don't we need to add the `reason` as well?
 
```python
from simpl.utils.importing import import_me_maybe

eventlet = import_me_maybe(
    'eventlet', reason="Module blah.blah requires eventlet for optimal "
                       "performance. Eventlet is not available so a slower"
                       "algorithm will be used.")  # pylint: disable=C0103
```

ziadsawalha:
I think this:
```python
from simpl.utils.importing import import_me_maybe
import_me_maybe('keyring')
```
will be unpythonic in it's implicit behavior. And it will throw many developers and static code analyzers off.
> Where the heck is `keyring` defined? Oh... magic...

This is more explicit and pythonic, IMO:
```python
from simpl.utils.importing import import_me_maybe
keyring = import_me_maybe('keyring')  # pylint: disable=C0103
```
but I believe the fair comparison is against:
```python
import warnings
try:
    import keyring
except ImportError:
    warnings.warn("'keyring' not installed. Config support for it disabled.")
    keyring = None  # pylint: disable=C0103
```

samstav:
Here's another (possibly bad) idea, thought I'd see what y'all think:

What if `import_me_maybe` had this signature

```python
def import_me_maybe(*modules):
    ...
```

and you could do things like this:

```python
# an embellished example
json = import_me_maybe('ujson', 'simplejson', 'json')
```

I think that is self-explanatory... you would get the `ujson` module if it was there, otherwise `simplejson` and finally `json`.

simpl/utils/importing.py
```diff
@@ -0,0 +1,85 @@
...
+    else:
+        importer_path = importer[1]
+    fq_path = importer_path.split(os.path.sep)
+    if 'simpl' in fq_path:
+        # find the rightmost instance of 'simpl' in path
+        begin = (len(fq_path) - 1) - fq_path[::-1].index('simpl')
+        import_string = '.'.join(fq_path[begin:-1] + [importer_module_name])
+    else:
+        import_string = importer_module_name or importer[1]
+    try:
```
samstav: Note to self.... lines 50-68 have no reason to be executed if the `ImportError` does not occur.

Pablosan:
After a conversation with @samstav I'd like to float a bit of an extremist idea: what if we say "`simpl` is highly opinionated about everything in `simpl`"? If we did that, I think the solution this PR presents becomes unnecessary. Let me explain...

If one wishes to use the (yet to be added) `simpl.server` module it is absolutely required that `bottle` and `eventlet` are installed, and we would specify this in `requirements/server.txt`. If one wishes to use the (yet to be added) `simpl.mongo` module it is absolutely required that `pymongo` is installed, and we would specify this in `requirements/mongo.txt`. If one tries to import either of these without `pip install`ing their associated requirements file they should expect to receive an `ImportError`.

This leaves the "optional" use case (e.g. `simpl.config` will use `keyring` if you have it installed but will also work just fine without it). If we go full extremist we would say "keyring is absolutely required" (and we could eliminate some code from `simpl.config`), but I think it would also be okay to just omit a requirements file because we have coded that module to treat a package as optional.

So, I think a possible guideline might be: `simpl` is highly opinionated about almost everything. When it chooses to relax the rules, it will do so quietly (without throwing exceptions). However, with what we are talking about including in `simpl` so far I'd be totally fine with: `simpl` is highly opinionated about _everything_.

ziadsawalha:
I like that approach, @Pablosan & @samstav . May I suggest a minor tweak: that we also be OK with raising the error later (not on import) if parts of the code have the possibility of being useful.

For example, I've implemented the CORS middleware in a [similar way](https://github.com/checkmate/simpl/pull/27/files#diff-2bf8a32a4898298a88f9196c73c69298R83), but it only fails when the `webob` import is actually used, not when the import fails. What that gives us is more frequently reusable shared code because the CORS module and classes are still available to be imported so that any constants or partial code can be used without causing an error. For example, let's say I want to subclass CORSMiddleware and re-implement \_\_call\_\_ without using WebOb. Another use case is if I want to access the cors.CORSMiddleware.default_headers constant in some other WSGI code. In both these examples the `webob` import is in fact not needed and should not block the module import.

Pablosan:
@ziadsawalha that makes sense to me. Should we add these guidelines to the README? It seems like we should document this someplace.

ziadsawalha:
How about a CONTRIBUTING.md?

Pablosan:
I'll work on getting our thoughts organized in a new CONTRIBUTING.md and submit a PR today.

@samstav should we close this PR without merging at this point?

samstav:
Sounds good @Pablosan. Thanks everyone for the discussion.

Pablosan:
The optional dependency guidelines have been captured in PR #47.

PR #19, opened 2015-05-29
WIP: server.py

samstav:
### Need to finish discussion and work here first: https://github.com/checkmate/simpl/pull/16

---------

#### todo

- [ ] finish design/work for conditional requirements: #16 
- [ ] finish work for generic middleware: #15
- [ ] tests

### DIFF
--- a/simpl/server.py
+++ b/simpl/server.py
@@ -1,0 +1,119 @@
+# Copyright (c) 2011-2015 Rackspace US, Inc.
+# All Rights Reserved.
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+"""Simpl bottle-based server.
+
+Runs with the following built-in middlewares by default:
+    StripPathMiddleware:
+        - strips/normalizes extra "/"s
+    CORSMiddleware
+        -
+
+"""
+import sys
+
+from simpl.utils import import_me_maybe
+
+bottle = import_me_maybe('bottle')
+eventlet = import_me_maybe('eventlet')
+
+_server_adapter = 'eventlet' if eventlet else 'wsgiref'
+
+
+def run(app=None,
+        server=_server_adapter,
+        host='127.0.0.1',
+        port=8080,
+        interval=1,
+        reloader=False,
+        quiet=False,
+        plugins=None,
+        debug=None,
+        **kwargs):
+    """Start a server instance. This method blocks until the server terminates.
+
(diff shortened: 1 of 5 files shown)


simpl/server.py
```diff
@@ -0,0 +1,119 @@
...
+        port=port,
+        interval=interval,
+        reloader=reloader,
+        quiet=quiet,
+        plugins=plugins,
+        debug=debug,
+        **kwargs)
+
+
+if __name__ == '__main__':
```
ziadsawalha: Do we need to wrap this in a function to make it callable from a setuptools entry_point or is there a way to have an entry_point execute a .py file directly?

simpl/server.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if not bottle:
+        sys.exit()
+
+    import argparse
+
+    from simpl.utils import cli
+
+    parser = cli.HelpfulParser(
```
ziadsawalha: Since simpl already has the config module, why would we not use that?

samstav: Now you're talking

simpl/server.py
samstav: I actually considered starting `simpl`'s CLI for this.... `simpl server --app ~/myapp/server.py --port 8888`

Let me see what this would look like...

```python
# simpl/__init__.py

# imagine everything else from simpl/__init__.py is here too

from simpl.utils import cli

PARSER = cli.HelpfulParser(prog=__title__)
SUBPARSERS = PARSER.add_subparsers(dest='_subparsers', title='simpl commands', ...)

def default_parser():
    return PARSER

def default_subparsers():
    return SUBPARSERS

```
and

```python
# simpl/server.py

from simpl import config

OPTIONS = [
    config.Option(
        '--app', '-a',
        help=("WSGI application to load by name. "
              "Ex: package.module gets the module"
              "    package.module:name gets the variable 'name'"
              "    package.module.func() calls func() and gets the result")
    ),
    config.Option(
        '--host', help="Server address to bind to, defaults to 127.0.0.1"
    ),
    config.Option(
        '--port', help="Server port to bind to"
    ),
    ...
]

SERVER_CONFIG = config.Config(options=OPTIONS)

...

if __name__ == '__main__':

    import simpl

    # not sure about unforseen complications with running
    # multiple argparsers, I think this should be OK
    SERVER_CONFIG.parse()

    def _run(args, **bottle_kwargs):
        run(app=args.app, server=args.server, host=args.host,
            port=args.port, reloader=args.reloader, **bottle_kwargs)

    subparsers = simpl.default_subparsers()
    server_parser = subparsers.add_parser(
        'server', help='Run a bottle microframework based wsgi server')
    server_parser.set_defaults(func=_run)

    args, extras = parser.parse_known_args()
    bottle_kwargs = {key.replace('-', '_'): val
                     for key, val in zip(extras[::2], extras[1::2])}
    
    args.func(args, **extras)



```
Thoughts ? There are definitely some kinks to work out here WRT usage/help output too....

ziadsawalha: What I like about this is that it provides a simple way to start an app, but also acts as a usage example that can easily be copied over and extended.

simpl/server.py
ziadsawalha: Nice.

samstav:
Supplanted by https://github.com/checkmate/simpl/pull/82

PR #20, opened 2015-06-11
Support for Python 3 and Style Checking

ziadsawalha:
So we can use it in patrick and juno...

Now that we have more than one Python 3 project (with the addition of Juno), we need to support both.


simpl/config.py
```diff
@@ -213,6 +213,7 @@ def __init__(self, *args, **kwargs):
         self._mutexgroup = None
 
     def __copy__(self):
+        """TODO(zns): add docstring."""
```
Pablosan: Is this just here to suppress PEP8 warnings? Why not just let the warnings continue until a proper docstring is added?

tests/test_config.py
```diff
@@ -161,7 +162,7 @@ def test_mutual_exclusion(self, mock_stderr):
         # for read_from
         keystring = 'this-is-a-private-key'
         strfile = tempfile.NamedTemporaryFile()
-        strfile.write('%s-written-to-file' % keystring)
+        strfile.write(('%s-written-to-file' % keystring).encode('ascii'))
```
Pablosan: Is ASCII encoding (reversion) necessary? Are there current filesystems that don't support unicode/UTF-8?

simpl/config.py
samstav: Magic (double-underscore methods) don't need a docstring right?

samstav:
FYI here's a other way to run builds/test against multiple python versions when using travisCI. My gut reaction was to offload this complexity to the CI service, but I also see the usefulness of letting tox do the work...

http://docs.travis-ci.com/user/languages/python/#Choosing-Python-versions-to-test-against

tests/test_config.py
ziadsawalha: This is just to support Python 2/3. I guess utf-8 could work.

ziadsawalha:
@samstav with tox you can run the tests locally the same way as with travis. I don't think that is possible with just the .travis.yml file, is it?

simpl/config.py
ziadsawalha: Yes. This was just for PEP-8. Any idea how to disable them from a config side? pyintrc can have the setting below, but I don't know how to do that for flake8:
```
# Don't require docstrings on tests or Python Special Methods
no-docstring-rgx=(__.*__|[tT]est.*|setUp|tearDown)
```

ziadsawalha:
@Pablosan @samstav - updated docsgtrings, changed ascii to utf-8.

samstav:
Thanks @ziadsawalha

PR #23, opened 2015-06-15
feat(logging): add logging module

ziadsawalha:
The simpl.logging module encapsulates some common
logging boilerplate. It also integrates with the
config module to support configuring logging using
command-line arguments, environment variables,
config files, and such.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,24 @@
 # simpl
-common libraries
+
+Common Python libraries for:
+
+- [Configuration](#config)
+- [Logging](#logging)
+
+## <a name="config"></a>Config
+
+Supports argparse-like configuration options with support for the following
+configuration methods:
+- command-line arguments
+- environment variables
+- keychain (OSX) and keyring (Linux)
+- ini/config files
+
+## <a name="logging"></a>Logging (simpl.log)
+
+Encapsulates logging boilerplate code to initialize logging using the
+[config](#config) module.
+
 
 ## release
 [![latest](https://img.shields.io/pypi/v/simpl.svg)](https://pypi.python.org/pypi/simpl)
--- a/simpl/config.py
+++ b/simpl/config.py
@@ -20,7 +20,6 @@ Configurable parser that will parse config files, environment variables,
 keyring, and command-line arguments.
 
 
-
 Example test.ini file:
 
     [defaults]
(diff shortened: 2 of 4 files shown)


samstav:
Do you think we should change the name of the new file to something other than `logging.py` ? That way, if one of the other modules in the directory does an `import logging` they don't accidentally get the wrong module?

simpl/logging.py
```diff
@@ -0,0 +1,180 @@
...
+                  action="store_true",
+                  help="turn up logging to DEBUG (default is INFO)"),
+    config.Option("-q", "--quiet",
+                  default=False,
+                  action="store_true",
+                  help="turn down logging to WARN (default is INFO)"),
+]
+
+
+def getLogger(name):  # pylint: disable=C0103
```
samstav: What about just pointing this function to `logging.getLogger` that way you get the `__doc__` and everything else?

```python
getLogger = logging.getLogger
```

ziadsawalha: +1
Should we `from loggin import *`?

Pablosan:
Regarding changing the module name: I can't decide whether I'm concerned about accidentally importing the wrong `logging` or not. @samstav are you concerned about usage of simpl `logging` within simpl itself? Even then, though, wouldn't you have to `from simpl import logging`?

Which use case am I missing that would make which logging is being used unclear?

samstav:
What are your thoughts of following `threadlocal.py`'s API with `default()` and `new()` instead of `current()` and `init()`

samstav:
@Pablosan Re: https://github.com/checkmate/simpl/pull/23#issuecomment-112189397

I was concerned about someone developing on simpl, as well as the modules that currently live in the `simpl` directory:  

https://github.com/checkmate/simpl/tree/master/simpl

For example, this line: https://github.com/checkmate/simpl/blob/60494927d7ff0c0ad12857acd01aa764d4d27fdb/simpl/git.py#L40

I think that would grab the file called `logging.py` in that directory rather than python's `logging` module

ziadsawalha:
@samstav Heeding the practice of not naming modules names that conflict with standard libraries sounds wise. What would be a better name for the file?

samstav:
Also, @ziadsawalha, `threadlocal.py` does not make use of `global`s, but seems like it could have a matching API.

I am not sure if calling a function, `init()`, should necessarily modify the module-level global `CURRENT_CONFIG`

samstav:
@ziadsawalha re: https://github.com/checkmate/simpl/pull/23#issuecomment-112191972

I am not sure.... I think we could just leave it as `logging.py`, especially if we include the `from logging import *`

ziadsawalha:
We could call it `log.py`

```python
from simpl import log

LOG = log.getLogger(__name__)
```

ziadsawalha:
@samstav - I will rename it to log.py. That feels like more in line with good/common practice.

But I’m still not sure about new() and default(). The paradigm is not quite the same. new() is not really creating a new config, it's initializing the global singleton with the set of passed-in options. What might clarify it is if I change current() to fail if init was not called first...

PR #25, opened 2015-06-16
Package info simplification and reorganization

samstav:
Puts all package information into `simpl/__about__.py` for an improved layout and cleaner  `simpl/__init__.py` and `setup.py` 

This PR bumps the version to `0.2.2` and I'll cut a release after.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,68 +14,54 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""simpl packaging and installation."""
+"""simpl packaging, installation, and package attributes."""
 
-import ast
-import re
+import os
 from setuptools import find_packages
 from setuptools import setup
 
 
-DEPENDENCIES = [
+src_dir = os.path.dirname(os.path.realpath(__file__))
+
+about = {}
+with open(os.path.join(src_dir, 'simpl', '__about__.py')) as abt:
+    exec(abt.read(), about)
+
+
+INSTALL_REQUIRES = [
     'six',
 ]
+
 TESTS_REQUIRE = [
     'mock',
 ]
 
+CLASSIFIERS = [
+    'Intended Audience :: Developers',
+    'License :: OSI Approved :: Apache Software License',
+    'Operating System :: OS Independent',
+    'Topic :: Software Development',
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 2',
+    'Programming Language :: Python :: 2.7',
+    'Programming Language :: Python :: 3.4',
+]
 
-def package_meta():
-    """Read __init__.py for global package metadata.
-
-    Do this without importing the package.
-    """
-    _version_re = re.compile(r'__version__\s+=\s+(.*)')
-    _url_re = re.compile(r'__url__\s+=\s+(.*)')
-    _license_re = re.compile(r'__license__\s+=\s+(.*)')
-
-    with open('simpl/__init__.py', 'rb') as simplinit:
-        initcontent = simplinit.read()
-        version = str(ast.literal_eval(_version_re.search(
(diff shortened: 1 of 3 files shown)


ziadsawalha:
What do we gain by creating `__about__.py`?

setup.py
```diff
@@ -14,68 +14,54 @@
...
 from setuptools import find_packages
 from setuptools import setup
 
 
-DEPENDENCIES = [
+src_dir = os.path.dirname(os.path.realpath(__file__))
+
+about = {}
+with open(os.path.join(src_dir, 'simpl', '__about__.py')) as abt:
+    exec(abt.read(), about)
```
ziadsawalha: Would you consider something like the below instead of `exec`?

```python

import ast
import os
import re

def package_meta():
    """Read __init__.py for global package metadata.

    Do this without importing the package.
    """
    _version_re = re.compile(r'__version__\s+=\s+(.*)')
    _url_re = re.compile(r'__url__\s+=\s+(.*)')
    _license_re = re.compile(r'__license__\s+=\s+(.*)')

    with open(os.path.join('simpl', '__init__.py', 'rb') as initfile:
        initcontent = initfile.read()
        version = str(ast.literal_eval(_version_re.search(
            initcontent.decode('utf-8')).group(1)))
        url = str(ast.literal_eval(_url_re.search(
            initcontent.decode('utf-8')).group(1)))
        licencia = str(ast.literal_eval(_license_re.search(
            initcontent.decode('utf-8')).group(1)))
    return {
        'version': version,
        'license': licencia,
        'url': url,
    }
```

samstav: You mean leave it the way it was?

samstav: I'd rather use `exec`, since its cleaner and more readable. I don't see any danger here....

samstav:
re: https://github.com/checkmate/simpl/pull/25#issuecomment-112533620

We gain the ability to eval/exec it without having to run any other code. Makes this line super simple: `exec(abt.read(), about)`

setup.py
Pablosan: I definitely prefer `exec` over `eval`.

Pablosan:
This looks very clean to me and the pattern is not without precedent (see projects [here](https://github.com/pyca)).

:+1: LGTM

ziadsawalha:
Sold.

setup.py
ziadsawalha: @Pablosan  Can you elaborate on why you prefer exec over eval?

I'm not sure eval is an option here. They both pretty much compile a string to code and execute it, so I might be missing something?

ziadsawalha: @samstav - I think literal_eval compiles/accepts such a small set of operations that it is considered safe.

However, you are only executing a file which gets loaded by our code anyway and you're not doing any string manipulation or incorporating unknown inputs, so the risk is small. Merged.

If I were to suggest a risk, it would be that setup.py is usually run at a higher privilege (often root) than when called during runtime. So if __about__.py were manipulated in any way, it would be run as root. While that's a stretch, its obscure things like that that make me shy away from exec where possible. [this is coming from the person who put exec statements in checkmate, mind you]

Pablosan: I considered changing my comment as it's an over-simplification. What I didn't like about the `ast.literal_eval` solution was the proliferation of regex'es. One of my oft-repeated quotes: solve a problem with regex, now you have two problems. Additionally, the decode -> regex match -> literal_eval muddles things.

Since @samstav was adding even more information, that meant more regexes and the code was already fairly obfuscated to my eyes.

The `exec` approach is crystal clear to me: read the stuff out of this file and stuff it in a dict. That works for me. :)

ziadsawalha: Thanks for the clarification. I'm not in love with the code I suggested above... but it avoided `exec`.

PR #28, opened 2015-06-23
add docs directory

wanghaven:

PR #29, opened 2015-06-23
Revert "add docs directory"

samstav:
Reverts checkmate/simpl#28


samstav:
We need to figure out why the PR tests didn't pass and re-do the PR

PR #33, opened 2015-06-24
added some basic formatting and sectioning

wanghaven:


samstav:
The rainbow pic is a little big. Can we shrink it? Or, what do you think about this pic?

prism

samstav:
Another suggestion would be to post some screenshots here that exhibit the changes made.

PR #34, opened 2015-06-24
update some docstrings in code and add some referencing

wanghaven:


samstav:
:clap: thanks @wanghaven

PR #37, opened 2015-06-26
--ini error conditions

samstav:
This change raises exceptions on these conditions:
    - an explicitly defined ini file does not exist
    - a config option found in a loaded ini file
      does not correspond to a `config.Option`

Add `SimplConfigException` and `SimplConfigUnknownOption`

`SimplConfigUnknownOption` will be thrown if an ini file is
explicitly specified, but there are options in it that have
no match, i.e. no corresponding `config.Option`

Previously, if these conditions occurred, nothing would happen. @Gifflen spent quite a long time trying to figure out why the options defined in his ini file were not getting picked up by simpl.

### DIFF
--- a/simpl/config.py
+++ b/simpl/config.py
@@ -359,6 +359,7 @@ class Config(collections.MutableMapping):
         """
         self._parser_kwargs = parser_kwargs or {}
         self._ini_paths = ini_paths or []
+        self._ini_paths = [normalized_path(x) for x in self._ini_paths]
         self._options = copy.copy(options) or []
         self._values = {option.name: option.default
                         for option in self._options}
@@ -387,11 +388,6 @@ class Config(collections.MutableMapping):
     def prog(self, value):
         """Set program name."""
         self._prog = value
-
-    @property
-    def default_ini(self):
-        """Default ini file name."""
-        return '%s.ini' % self.prog
 
     def _metaconfigure(self, argv=None):
         """Initialize metaconfig for provisioning self."""
@@ -541,10 +537,6 @@ class Config(collections.MutableMapping):
         results = {}
         self.ini_config = configparser.SafeConfigParser()
 
-        if os.path.isfile(self.default_ini) and (
-                self.default_ini not in self._ini_paths):
-            self._ini_paths.append(self.default_ini)
-
         parser_errors = (configparser.NoOptionError,
                          configparser.NoSectionError)
 
@@ -552,9 +544,6 @@ class Config(collections.MutableMapping):
         # check that explicitly defined ini paths exist
         for pth in inipaths:
             if not os.path.isfile(pth):
-                inipaths.remove(pth)
-                if pth == self.default_ini:
-                    continue
                 raise OSError(errno.ENOENT, 'No such file or directory', pth)
 
         read_ok = self.ini_config.read(inipaths)
@@ -666,6 +655,15 @@ class Config(collections.MutableMapping):
             '%s=%s' % (k, v) for k, v in self.items()])
 
 
+def normalized_path(value):
+    """Normalize and expand a shorthand or relative path."""
+    if not value:
+        return
+    norm = os.path.normpath(value)
+    norm = os.path.abspath(os.path.expanduser(norm))
+    return norm
+
+
 class MetaConfig(Config):
 
     """A config class used by Config to find and evaluate config sources.
@@ -747,10 +745,8 @@ class MetaConfig(Config):
 
     options = [
         Option('--ini', metavar='PATH',
-               help=('Source some or all of the options from this ini file. '
-                     'Defaults to <program_name>.ini in your '
-                     'current working directory.'),
-               default='%s.ini' % sys.argv[0],
+               help=('Source some or all of the options from this ini file.'),
+               type=normalized_path,
                group=option_group, group_description=option_description),
     ]
 
@@ -782,15 +778,6 @@ def read_from(value):
     with open(path, 'r') as reader:
         read = reader.read()
     return read
-
-
-def normalized_path(value):
-    """Normalize and expand a shorthand or relative path."""
-    if not value:
-        return
-    norm = os.path.normpath(value)
-    norm = os.path.abspath(os.path.expanduser(norm))
-    return norm
 
 
 def comma_separated_strings(value):
--- a/simpl/config.py
+++ b/simpl/config.py
@@ -165,9 +165,12 @@ from __future__ import print_function
 import argparse
 import collections
 import copy
+import errno
 import logging
 import os
 import sys
+
+import simpl.exceptions
 
 try:
     import keyring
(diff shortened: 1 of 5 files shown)


simpl/exceptions.py
```diff
@@ -107,3 +107,13 @@ def __str__(self):
...
+
+
+class SimplConfigException(SimplException):
+
+    """Errors raised by simpl/config."""
+
+
+class SimplConfigUnusedIniOption(SimplConfigException):
+
+    """An option defined in the specified ini file has no match."""
```
samstav: self-critique... This exception name is bad.

simpl/config.py
```diff
@@ -526,7 +525,7 @@ def get_defaults(self):
         parsed, _ = parser.parse_known_args([])
         return vars(parsed)
 
-    def parse_ini(self, paths=None, namespace=None):
+    def parse_ini(self, paths=None, namespace=None, permissive=False):
```
samstav: I added the permissive argument, but there's not really any easy way to use it, unless you were to subclass Config and override the `load_options` method. 

I assumed (and agree) that throwing an error upon encountering "loose" options might not be the desired behavior for some. For the record, this works very well for our use cases, and would have alerted us to problems sooner, although I would argue that it would be nice if you could turn this feature on more easily, i.e. the `permissive` feature of our ini parsing.

simpl/exceptions.py
ziadsawalha:
```
$python -f
Unknown option: -f
```

How about `SimplConfigUnknownOption` and we use it for arguments as well as ini options?

simpl/config.py
ziadsawalha: Supporting different parsing between ini files and command arguments will make the interface more complicated than needed. So I think if we add it we should add `permissive=` kwarg (instead of `permissive_ini=, permissive_args, etc..`
But... having said that... we can leave this out until it is required. As you say, we don't need it for our use cases, so we could skip it until (if ever) someone asks for it.
In other words, the code as is is good, IMO

simpl/config.py
```diff
@@ -165,10 +165,13 @@ def main(argv):
 import argparse
 import collections
 import copy
+import errno
 import logging
 import os
 import sys
 
+from simpl import exceptions as simpl_exceptions
```
ziadsawalha: It would be nice if we came up with a convention for importing exceptions.

Some things we've discussed before that maybe we can add to a contributing.md or in the docstring for simpl/exceptions.py:
- A fully-named exception is useful when exceptions are caught and bubble up (ex. SimplConfigError is easier to track than simpl.ConfigError if it bubbles out of simpl)
- "exceptions" is a long word. One of our projects puts exceptions in `exc.py`; I think that's a good idea.
- valid reasons for renaming imports are to avoid conflicts or to shorten very long names.
simple_exceptions is no shorter than simpl.exceptions. So why not just:
```python
import simple.exceptions

...
raise simpl.exceptions.SimplConfigUnusedIniOption()
```

tests/test_config.py
```diff
@@ -42,6 +44,22 @@ def test_format_comma_separated_pairs(self):
 
 
 class TestConfig(unittest.TestCase):
+
+    def setUp(self):
```
ziadsawalha: Should we PEP test files? Maybe not full analysis, but at least PEP-8?

tests/test_git.py
```diff
@@ -283,15 +283,25 @@ def test_clone_brings_branches(self):
...
-        self.assertIn({
-            'commit': self.repo.head,
-            'message': 'On master: stash message',
-            'branch': '(detached from %s)' % self.repo.head[:7]
-        }, branches)
+        selected = None
+        for b in branches:
+            if b['commit'] == self.repo.head:
+                selected = b
+                break
```
ziadsawalha: No changes required, just musings on python...

I wonder/wish if there was a simpler pattern for this in python. Its such a common occurrence.

This works but its not inconclusively simpler or clearer.
```python
        selected = next((b for b in branches if b['commit'] == self.repo.head), None)
```

tests/test_git.py
```diff
@@ -283,15 +283,25 @@ def test_clone_brings_branches(self):
...
+        self.assertIsNotNone(selected)
+        self.assertEqual(selected['commit'], self.repo.head)
+        self.assertEqual(selected['message'], 'On master: stash message')
+        try:
+            fmtr = {'cmt': self.repo.head[:7]}
+            self.assertEqual(
+                selected['branch'], '(detached from {cmt})'.format(**fmtr))
+        except AssertionError as e:
```
ziadsawalha: Are these different possible responses based on the version of git?

simpl/config.py
```diff
@@ -538,13 +537,24 @@ def parse_ini(self, paths=None, namespace=None):
...
+        for pth in inipaths:
+            if not os.path.isfile(pth):
+                raise OSError(errno.ENOENT, 'No such file or directory', pth)
+
+        read_ok = self.ini_config.read(inipaths)
+        assert read_ok == inipaths
+        dicts = (list(self.ini_config._sections.values()) +
+                 [self.ini_config.defaults()])
+        ini_options = copy.deepcopy(
+            {k: v for d in dicts for k, v in d.items() if k != '__name__'})
```
samstav: Change this to a `set` of keys rather than a dict. You don't need the values.

simpl/config.py
samstav: :+1: Thanks.

tests/test_git.py
samstav: > If they had called it `elifnobreak`, nobody would be confused.

- Raymond Hettinger

-----------

https://youtu.be/OSGv2VnC0go

tests/test_git.py
samstav: Yes, I'll add a comment.

samstav: https://github.com/samstav/simpl/commit/03068c2da07ce5b26306d8922883fb2e7e49bd7e

samstav:
Thanks for the thorough review @ziadsawalha. I think I've addressed all comments, let me know.

simpl/config.py
```diff
@@ -165,10 +165,13 @@ def main(argv):
 import argparse
 import collections
 import copy
+import errno
 import logging
 import os
 import sys
 
+import simpl.exceptions
```
ziadsawalha: Much cleaner, IMO. Thanks.

tests/test_config.py
```diff
@@ -42,6 +44,13 @@ def test_format_comma_separated_pairs(self):
 
 
 class TestConfig(unittest.TestCase):
+
+    def get_tempfile(self, *args, **kwargs):
+
```
ziadsawalha: Blank line after `def` contradicts PEP-8.

ziadsawalha:
:+1: - last two comments are minor.

PR #38, opened 2015-06-29
feat(rest): add REST API Helpers v1

ziadsawalha:
This adds the first of the `rest.py` helpers that handle common bottle needs when implementing our APIs. Needs such as parsing the body and query params.

### DIFF
--- a/simpl/rest.py
+++ b/simpl/rest.py
@@ -1,0 +1,55 @@
+# Copyright (c) 2011-2015 Rackspace US, Inc.
+# All Rights Reserved.
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+"""REST-ful API Utilites."""
+
+import bottle
+
+
+def body(schema=None, types=None, required=False, default=None):
+    """Decorator to parse and validate API body.
+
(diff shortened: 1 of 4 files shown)


simpl/rest.py
```diff
@@ -0,0 +1,55 @@
...
+    if not types:
+        types = ['application/json']
+    if not all('json' in t for t in types):
+        raise NotImplementedError("Only 'json' body supported.")
+
+    def wrap(fxn):
+        """Return a decorated callable."""
+        def wrapped(*args):
+            """Callable to called when the decorated function is called."""
+            data = bottle.request.json
```
samstav: Is this where you'd like to have a `RuntimeError` ?

https://github.com/checkmate/simpl/pull/16#issuecomment-114304925

tests/test_rest.py
```diff
@@ -0,0 +1,78 @@
...
+        route()
+        mock_handler.assert_called_once_with(int(data))
+
+    @mock.patch.object(rest.bottle, 'request')
+    def test_schema_fail(self, mock_request):
+        """Test schema is enforced."""
+        mock_request.json = 'ALPHA'
+        mock_handler = mock.Mock()
+        route = rest.body(schema=int)(mock_handler)
+        with self.assertRaises(rest.bottle.HTTPError):
```
samstav: This is an interesting approach to importing `HTTPError`... I think convention would have you import `bottle` in this test file ?

simpl/rest.py
ziadsawalha: I don't see anything useful in the module that could be used without bottle right now, so letting it fail on import if bottle is not there feels appropriate.

If we do eventually add code that can be used without bottle we could then add the conditional import and runtime check. If we add it here, it will run on each HTTP call to this route. We could add it one scope above so it only runs on decoration (once at load). I don't know what would be better; I guess it would depend on the code that comes in.

tests/test_rest.py
ziadsawalha: True. Referring to `bottle` in the `rest` when patching is correct (like in L48), but this here is an unnecessary shortcut. I'll fix it.

samstav:
I don't love that this is coupled to bottle, but I'm going to merge it since I don't know how else you'd have access to the request instance...

PR #40, opened 2015-06-29
feat(dicts): utilities for manipulating dicts

ziadsawalha:
This pulls some of the python dict manipulation code from utils.py.

### DIFF
--- a/tests/test_dicts.py
+++ b/tests/test_dicts.py
@@ -1,0 +1,440 @@
+# Copyright 2013-2015 Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for :mod:`dicts`"""
+
+import copy
+import re
+import unittest
+
+import mock
+import six
+
+from simpl.incubator import dicts
+
+
+class TestSplitMergeDicts(unittest.TestCase):
+
+    """Tests for split/merge dicts in :mod:`dicts`."""
+
+    def test_split_dict_simple(self):
+        fxn = dicts.split_dict
+        self.assertEqual(fxn({}), ({}, None))
+        combined = {
+            'innocuous': 'Hello!',
+            'password': 'secret',
+        }
+        innocuous = {'innocuous': 'Hello!'}
+        secret = {'password': 'secret'}
+        original = combined.copy()
+        self.assertEqual(fxn(combined, filter_keys=[]), (combined, None))
+        self.assertEqual(fxn(combined, ['password']), (innocuous, secret))
+        self.assertDictEqual(combined, original)
+
+    def test_split_dict_works_with_None_keys(self):
+        filter_keys = [re.compile('quux')]
+        data = {None: 'foobar'}
+        expected = (data, None)
+        self.assertEqual(expected,
+                         dicts.split_dict(data, filter_keys))
+
+    def test_extract_data_expression_as_filter(self):
+        data = {
+            "employee": {
+                "name": "Bob",
+                "title": "Mr.",
+                "public_key": "rsa public key",
+                "private_key": "a private key",
+                "password": "password",
+                "position": "left"
+            },
+            "server": {
+                "access": {
+                    "rootpassword": "password",
+                    "server_privatekey": "private_key",
+                    "server_public_key": "public_key"
+                },
+                "private_ip": "123.45.67.89",
+                "public_ip": "127.0.0.1",
+                "host_name": "server1"
+            },
+            "safe_val": "hithere",
+            "secret_value": "Immasecret"
+        }
+
+        safe = {
+            "employee": {
+                "name": "Bob",
+                "title": "Mr.",
+                "public_key": "rsa public key",
+                "position": "left"
+            },
+            "server": {
+                "access": {
+                    "server_public_key": "public_key"
+                },
+                "private_ip": "123.45.67.89",
+                "public_ip": "127.0.0.1",
+                "host_name": "server1"
+            },
+            "safe_val": "hithere",
+        }
+
+        secret = {
+            "employee": {
+                "private_key": "a private key",
+                "password": "password",
+            },
+            "server": {
+                "access": {
+                    "rootpassword": "password",
+                    "server_privatekey": "private_key",
+                }
+            },
+            "secret_value": "Immasecret"
+        }
+
+        original_dict = copy.deepcopy(data)
+        secret_keys = ["secret_value", re.compile("password"),
+                       re.compile("priv(?:ate)?[-_ ]?key$")]
+        body, hidden = dicts.split_dict(data, secret_keys)
+        self.assertDictEqual(body, safe)
+        self.assertDictEqual(secret, hidden)
+        dicts.merge_dictionary(body, hidden)
+        self.assertDictEqual(original_dict, body)
+
+    def test_split_dict_complex(self):
+        fxn = dicts.split_dict
+        self.assertEqual(fxn({}), ({}, None))
+        combined = {
+            'innocuous': {
+                'names': ['Tom', 'Richard', 'Harry']
+            },
+            'data': {
+                'credentials': [{'password': 'secret', 'username': 'joe'}],
+                'id': 1000,
+                'list_with_only_cred_objects': [{'password': 'secret'}],
+                'list_with_some_cred_objects': [
+                    {
+                        'password': 'secret',
+                        'type': 'password',
+                    },
+                    'scalar',
+                    {'name': 'joe'}
+                ]
+            }
+        }
+        innocuous = {
+            'innocuous': {
+                'names': ['Tom', 'Richard', 'Harry']
+            },
+            'data': {
+                'id': 1000,
+                'list_with_some_cred_objects': [
+                    {
+                        'type': 'password'
+                    },
+                    'scalar',
+                    {'name': 'joe'}
+                ]
+            }
+        }
+        secret = {
+            'data': {
+                'credentials': [{'password': 'secret', 'username': 'joe'}],
+                'list_with_only_cred_objects': [{'password': 'secret'}],
+                'list_with_some_cred_objects': [
+                    {
+                        'password': 'secret'
+                    },
+                    None,
+                    {}
+                ]
+            }
+        }
+        original = combined.copy()
+        not_secret, is_secret = fxn(combined, [])
+        self.assertDictEqual(not_secret, combined)
+        self.assertIsNone(is_secret)
+
+        not_secret, is_secret = fxn(combined, ['credentials', 'password'])
+        self.assertDictEqual(not_secret, innocuous)
+        self.assertDictEqual(is_secret, secret)
+        self.assertDictEqual(combined, original)
+
+        merged = dicts.merge_dictionary(innocuous, secret)
+        self.assertDictEqual(original, merged)
+
+    def test_extract_and_merge(self):
+        fxn = dicts.split_dict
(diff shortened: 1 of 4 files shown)


samstav:
I wrote this implementation in waldo a while back, not realizing that there was already code in the repo that did the same thing:

https://github.com/rackerlabs/waldo/blob/ac1f761876f858b68be6eb5e1dbc795c8474d85c/waldo/utils.py#L173-L195

While I have no doubts about either implementation, this seems like the kind of code we should avoid using in general. Something about it gives me a hunch that something *else* went wrong, further upstream, or in a library we don't own, or a bad design decision was made....

What do we need this for? Am I wary without reason?

simpl/dicts.py
```diff
@@ -0,0 +1,133 @@
...
+    :param source: a dict to read data from
+    :param path: a key or path to a key (path is delimited by `separator`)
+    :keyword separator: the separator used in the path (ex. Could be "." for a
+        json/mongodb type of value)
+    """
+    parts = path.strip(separator).split(separator)
+    current = source
+    for part in parts[:-1]:
+        if part not in current:
+            return
```
larsbutler: Please add a test which exercises this `return` case. It seems a bit like an error/edge condition.

This behavior should also be document in the docstring: if a path does not exist, return None (as opposed to a `KeyError`).

simpl/dicts.py
```diff
@@ -0,0 +1,133 @@
...
+    :param source: a dict to read data from
+    :param path: a key or path to a key (path is delimited by `separator`)
+    :keyword separator: the separator used in the path (ex. Could be "." for a
+        json/mongodb type of value)
+    """
+    if path == separator and isinstance(source, dict):
+        return True
+    parts = path.strip(separator).split(separator)
+    if not parts:
+        return False
```
larsbutler: The case at line 68 (`return False`) does not have any tests.

simpl/dicts.py
```diff
@@ -0,0 +1,133 @@
...
+                    merge_lists(dest, source, extend_lists=extend_lists)
+                else:
+                    current_dst[key] = source
+    return dst
+
+
+def merge_lists(dest, source, extend_lists=False):
+    """Recursively merge two lists.
+
+    This applies merge_dictionary if any of the entries are dicts.
```
larsbutler: Can you explain the semantics of `extend_lists` in the docstring? Looking at the method signature, I had an idea about what this parameter means, but then when I looked at the code, my assumption was completely wrong. Looking at the test code in order to understand, I'm even more confused.

Some simple doctest examples would be ideal, to show the difference in behavior when `extend_lists` is `False` vs. when it is `True`.

simpl/dicts.py
```diff
@@ -0,0 +1,133 @@
...
+    Note: This updates dest and returns it.
+    """
+    if not source:
+        return
+    if not extend_lists:
+        # Make them the same size
+        left = dest
+        right = source[:]
+        if len(dest) > len(source):
+            right.extend([None for _ in range(len(dest) - len(source))])
```
larsbutler: Line 117 (`right.extend...`) is missing test coverage.

tests/test_dicts.py
```diff
@@ -0,0 +1,259 @@
...
+        self.assertEqual(result['l'], [[], [{'s': 1, 't': 8}]])
+
+    def test_merge_lists(self):
+        dst = [[], [2], [None, 4]]
+        src = [[1], [], [3, None]]
+        result = dicts.merge_lists(dst, src)
+        self.assertIsInstance(result, list)
+        self.assertEqual(result[0], [1])
+        self.assertEqual(result[1], [2])
+        self.assertEqual(result[2], [3, 4], "Found: %s" % result[2])
```
larsbutler: Why are the `None` values dropped here?

tests/test_dicts.py
```diff
@@ -0,0 +1,259 @@
...
+            c=[1],  # add new list
+            d=[None, None, 'c'],
+            e=[1, 2]
+        )
+        result = dicts.merge_dictionary(dst, src, extend_lists=True)
+        self.assertIsInstance(result, dict)
+        self.assertEqual(result['a'], [1])
+        self.assertEqual(result['b'], [1, 2])
+        self.assertEqual(result['c'], [1])
+        self.assertEqual(result['d'], ['a', 'b', None, None, 'c'])
```
larsbutler: And why are the not dropped here? (See https://github.com/checkmate/simpl/pull/40/files#r33656443.)

simpl/dicts.py
```diff
@@ -0,0 +1,133 @@
...
+    for part in parts:
+        if not isinstance(current, dict):
+            return False
+        if part not in current:
+            return False
+        current = current[part]
+    return True
+
+
+def merge_dictionary(dst, src, extend_lists=False):
```
larsbutler: The semantics of `merge_dictionary` are really strange; I'm not even sure how I could use this. Given the examples below, if I had to "merge" two dictionaries, I can't comprehend what I would expect to happen.

```
>>> from simpl import dicts
>>> d1 = dict(a=[1,2]); d2=dict(a=[2,1]); dicts.merge_dictionary(d1, d2, extend_lists=False)
{'a': [2, 1]}
>>> d1 = dict(a=[1,2]); d2=dict(a=[2,1]); dicts.merge_dictionary(d1, d2, extend_lists=True)
{'a': [1, 2]}
>>> d1 = dict(a=[1,2]); d2=dict(a=[2,1]); dicts.merge_dictionary(d2, d1, extend_lists=False)
{'a': [1, 2]}
>>> d1 = dict(a=[1,2]); d2=dict(a=[2,1]); dicts.merge_dictionary(d2, d1, extend_lists=True)
{'a': [2, 1]}
>>> d1 = dict(a=[1,None,2]); d2=dict(a=[None,2,1]); dicts.merge_dictionary(d1, d2, extend_lists=False)
{'a': [1, 2, 1]}
>>> d1 = dict(a=[1,None,2]); d2=dict(a=[None,2,1]); dicts.merge_dictionary(d1, d2, extend_lists=True)
{'a': [1, None, 2]}
>>> d1 = dict(a=[1,None,2]); d2=dict(a=[None,2,1]); dicts.merge_dictionary(d2, d1, extend_lists=False)
{'a': [1, 2, 2]}
>>> d1 = dict(a=[1,None,2]); d2=dict(a=[None,2,1]); dicts.merge_dictionary(d2, d1, extend_lists=True)
{'a': [None, 2, 1]}
>>> 
```

Can you explain this behavior? It almost seems like the lists are behaving like some kind ordered sets--sometimes. Sometimes the `dst` dominates the order and contents of the list, sometimes `None` values are dropped.

Does `src` override `dst` if a key in a nested dict exists in both, but with a different value?

```
>>> d1 = dict(a=dict(foo='baz')); d2=dict(a=dict(foo='bar')); dicts.merge_dictionary(d2, d1, extend_lists=False)
{'a': {'foo': 'baz'}}
>>> d1 = dict(a=dict(foo='baz')); d2=dict(a=dict(foo='bar')); dicts.merge_dictionary(d1, d2, extend_lists=False)
{'a': {'foo': 'bar'}}
>>> 
```

simpl/dicts.py
```diff
@@ -0,0 +1,133 @@
...
+        for index, value in enumerate(left):
+            if value is None and right[index] is not None:
+                dest[index] = right[index]
+            elif isinstance(value, dict) and isinstance(right[index], dict):
+                merge_dictionary(dest[index], source[index],
+                                 extend_lists=extend_lists)
+            elif isinstance(value, list):
+                merge_lists(value, right[index])
+            elif right[index] is not None:
+                dest[index] = right[index]
```
larsbutler: Line 130 (`elif right[index] is not None:` case) is not covered by tests.

ziadsawalha:
@samstav WRT https://github.com/checkmate/simpl/pull/40#issuecomment-116909976, I think these are different uses. We probably can consolidate and refactor some of these.
I've update the PR to address questions about the uses for this, but in the end this is code that exists in at least three of our projects and I need it in order to port code for my deliverable this sprint.

larsbutler:
> ... this is code that exists in at least three of our projects and I need it in order to port code ...

I don't buy the argument that, just because the code exists in other repos, we should propagate it to more repos. If the code has questionable design, unclear use cases, or other issues, we should pause and rethink things.

> ... my deliverable this sprint.

What is the deliverable exactly?

larsbutler:
I'm experimenting with those code more to try and figure out how it can be useful, and I'm still scratching my head. Can you give a real-world example of two dictionaries (which contain lists) that would be merged in this way?

Pablosan:
I believe our biggest risk with simpl is propagating poor design decisions we've made in the past. I'm not sure this code qualifies as poor design, but the fact that two devs so far are questioning it give me pause.

@ziadsawalha given the concerns posed so far I would strongly prefer to slow down, do some more research, and make sure everyone is confident in the approach but I understand that blocks you elsewhere. May I suggest duplicating the bare minimum amount of code needed in that project to unblock you?

Meanwhile I'm going to dig into this problem space. I see this PR as wanting to solve the following challenges:
  - Getting a nested value from a dict when we are not guaranteed that all of the keys in the path will be there
  - Setting a nested value in a dict, adding any missing keys in the path
  - Checking to see if a nested value exists in a dict
  - Dealing with dict partials (e.g. pulling a chunk of a dict out to deal with separately)
  - Safe, deep dict merges

There may be other utilities we could dream up, but those are the ones I see in the existing code. Looking at that list my first thought is "certainly these are solved problems." I'll report back once I've done my research.

tests/test_dicts.py
ziadsawalha: I've updated the PR with [this](https://github.com/checkmate/simpl/pull/40/files#diff-029888ac0a1327ac5d39627e785ea1ddR80) which documents a use case and explains the use of `None` values.

tests/test_dicts.py
ziadsawalha: Because `extend_lists` is set to `True`, so lists are not treated as positional overlays.

Pablosan:
@ziadsawalha and I just finished a good conversation that helped us identify competing demands we believe are being placed on `simpl`. I'll outline them briefly here, but I have also begun a new PR to add CONTRIBUTING.md that I hope will help us clarify what belongs in `simpl` and how it gets there.

The two competing demands:
  - `simpl` was dreamt up to solve a very basic problem: duplicate code. We found ourselves solving similar problems across multiple projects, many times copying code from one project to the next and then continuing to evolve the code there. The copy/paste approach has long been labeled an anti-pattern with many disadvantages. `simpl` is an attempt to eliminate those problems.
  - `simpl` is a package on which several projects already depend with hopes that even more projects will use it in the future. This puts a good deal of pressure on `simpl` to be "correct" (for some definition of "correct"), and we have even stated as a team that we want more rigor in our PR reviews to avoid propagating usage of bad patterns (for some definition of "bad").

The second demand tends to add friction to the process of getting new things into `simpl`, and I believe that friction has value. However, we are also often working on deadlines in other projects where we are trying to avoid the copy/paste approach that `simpl` was meant to solve. This PR definitely falls into that category. So how do we reconcile the two?

I have a proposed solution, and would ask that we continue this discussion on PR #47. Hopefully we can gain consensus and implement the solution on this PR in short order.

ziadsawalha:
Updated to go to `simpl.incubating.dicts`. This code exists and is in use in four of our projects. This PR attempts to deduplicate it and improve it slightly.

Pablosan:
I believe we should remove this code in favor of what appears to be an incredibly robust implementation: [jmespath](http://jmespath.org/). They have a Python implementation [here](https://pypi.python.org/pypi/jmespath/0.7.1). It appears to do most, if not all of what we need.

ziadsawalha:
@Pablosan - as noted when we merged this in, this code is here because it is duplicated in many of our projects.
I have not looked at jmespath, but it sounds great (more powrful API, less code for us to maintain), but from what I've read (very little so far, admittedly) I think it replaces only some of this code - the path functions, but not the deep merge.

Pablosan:
@ziadsawalha are any of those other projects using this code yet? If it's just being used in Juno, then I think I'd rather remove it from `simpl` to avoid the maintenance overhead.

Pablosan:
I have opened Issue #56 to further facilitate this discussion. I have toned down the recommendation somewhat (changed from "remove `dicts`" to "prefer `jmespath`").

ziadsawalha:
Thanks, @Pablosan - I am much more in favor of "prefer `jmespath`"

PR #41, opened 2015-06-30
fix(pkg): dont import git module in __init__

samstav:


simpl/__init__.py
```diff
@@ -17,7 +17,6 @@
 import os
 
 from simpl import config  # noqa
```
ziadsawalha: Why not remove this one too?

PR #42, opened 2015-06-30
fix(git import)

Pablosan:
Remove `from simpl import git.py` from `simpl/__init__.py` so the git
module warnings don't occur even when it is not being used.

### DIFF
--- a/simpl/__init__.py
+++ b/simpl/__init__.py
@@ -16,8 +16,6 @@
 
 import os
 
-from simpl import config  # noqa
-from simpl import git  # noqa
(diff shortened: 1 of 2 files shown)


simpl/__init__.py
```diff
@@ -17,7 +17,6 @@
 import os
 
 from simpl import config  # noqa
```
samstav: @ziadsawalha mentioned we can go ahead and remove this line too

Pablosan: @samstav thanks! Fix incoming...

PR #44, opened 2015-06-30
feat(rest): add `paginated` decorator

ziadsawalha:
An opinionated, paginated response formatter.

### DIFF
--- a/simpl/rest.py
+++ b/simpl/rest.py
@@ -14,7 +14,11 @@
 
 """REST-ful API Utilites."""
 
+import functools
+
 import bottle
+
+MAX_PAGE_SIZE = 10000000
 
 
 def body(schema=None, types=None, required=False, default=None):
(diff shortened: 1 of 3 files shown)


simpl/rest.py
```diff
@@ -53,3 +57,142 @@ def wrapped(*args):
...
+
+    for content_type in HANDLERS:
+        if content_type in accept:
+            return HANDLERS[content_type](data, request, response)
+
+    # Use default
+    return HANDLERS['default'](data, request, response)
+
+
+def write_json(data, request, response):
```
samstav: `request` is not used. Also might be worth considering whether the function is needed if it is only going go be 2 lines.

simpl/rest.py
```diff
@@ -53,3 +57,142 @@ def wrapped(*args):
...
+    response.set_header('Content-Type', 'application/json')
+    return to_json(data)
+
+
+def to_json(data):
+    """Write out python object to JSON.
+
+    Includes special handling for custom objects derived from MutableMapping.
+    """
+    if isinstance(data, collections.MutableMapping) and hasattr(data, 'dumps'):
```
samstav: I'm thrown off by this... a dict-like object with a `dumps` method? I can't think of an object that meets this criteria.

simpl/rest.py
```diff
@@ -53,3 +57,142 @@ def wrapped(*args):
...
+        return data.dumps(indent=4)
+    return json.dumps(data, indent=4)
+
+
+HANDLERS = {
+    'application/json': write_json,
+    'default': write_json
+}
```
samstav: What do you think about implementing something generic (multiple handlers) once we really need it and we do json only for now? Might be able to simplify this PR.

simpl/rest.py
```diff
@@ -53,3 +57,142 @@ def wrapped(*args):
...
+
+def paginated(uripath):
+    """Decorator that adds pagination information to the response header.
+
+    Opinionated assumptions:
+    - body has data items in `data` or `results`
+    - body has a `collection-count` value
+
+    Headers returned include:
+        Content-Range: ... per HTTP RFC and 206 response
```
samstav: RFC 7233 ? 

https://tools.ietf.org/html/rfc7233

samstav:
@ziadsawalha  If you get a chance, could you write a description of the logic in english? You might include assumptions, variations, or logical conditions for certain status codes, and circumstances you know would throw errors? That would help me with the review.

larsbutler:
A lot of the code in `rest.py` has no test coverage:

```
simpl.config              347     56    84%   259, 264-265, 330-333, 339, 376-378, 390, 396, 425, 429, 433, 437, 444, 487, 510-511, 564-568, 574-575, 591-597, 624-625, 632, 634-637, 648-649, 654, 661, 776, 800, 811-816, 835-842, 847-864, 867
simpl.git                 299     12    96%   139, 199-200, 281, 333, 431, 541, 728-732
simpl.log                  49     35    29%   87-94, 104-111, 124-133, 139-153, 165-169, 174-177
simpl.rest                 88     32    64%   39, 54, 69-76, 81-82, 90-92, 114-135, 156, 170-172, 178-180, 194
simpl.secrets              12      5    58%   32-36
TOTAL                     950    154    84%   
Ran 95 tests in 6.372s
```

simpl/rest.py
ziadsawalha: We also had `write_yaml`and other HANDLERs and some used `request`, so this is more of a signature. This is a copy/paste, not new code, but I should document how this works and how to extend it. I'll do that...

simpl/rest.py
ziadsawalha: ExtensibleDict or an implementation of collections.MutableMapping. Again, copy/paste that I will document better.

simpl/rest.py
ziadsawalha: Than we wouldn't be able to use this in Waldo and Checkmate...

simpl/rest.py
ziadsawalha: Yes. Section 4.2 specifically: https://tools.ietf.org/html/rfc7233#section-4.2

larsbutler:
> If you get a chance, could you write a description of the logic in english? You might include assumptions, variations, or logical conditions for certain status codes, and circumstances you know would throw errors? That would help me with the review.

+1

simpl/rest.py
ziadsawalha: I removed the response parsing from this PR and limited it to strict pagination.

ziadsawalha:
Added more docs to the pagination logic.

simpl/rest.py
```diff
@@ -53,3 +57,156 @@ def wrapped(*args):
...
+    It adds link headers based on RFC 5988 (by ex-Racker Mark Nottingham) -
+    https://tools.ietf.org/html/rfc5988#section-5.5.
+
+    Future changes to the RFCs or common API concepts, such as adding links to
+    the response body, may be implemented here to consistently handle
+    pagination across all projects using simpl pagination for their APIs.
+
+    Opinionated assumptions:
+    - body has data items under a `data` or `results` key.
+    - the `len()` of that data represents the number of records being returned
```
larsbutler: Does that mean that the value for `data` should be a list? This is an important detail for the APIs we are implementing.

ziadsawalha: No, it can be a list or dict (as long as `len()` returns the number of records returned in the call). We're returning `data: [...]` in some projects, which is a slight departure from `results: {...}` in other projects. It might even differ per call.

PR #47, opened 2015-07-01
chore(codify contribution guidelines)

Pablosan:
I will be updating this PR frequently over the next few hours, but want it to be visible as soon as possible so that the conversation can proceed apace.

This is our first attempt at documenting `simpl`'s raison d'être and guidelines for contributing to the project.

### DIFF
--- a/CONRTIBUTING.md
+++ b/CONRTIBUTING.md
@@ -1,0 +1,84 @@
+# Guidelines for Contributing to `simpl`
+
+`simpl` is a collection of common, recurring patterns we have found useful
+while building many REST API's.
+
+`simpl` is highly opinionated. You won't find generic implementations here. We
+stick to certain staples: [bottle](http://bottlepy.org/docs/dev/index.html),
+[mongodb](https://www.mongodb.org/), [redis](http://redis.io/),
+[git](https://git-scm.com/). Our patterns and modules are unabashedly coupled
+to these tools.
+
+While we make every effort to make `simpl` a set of top-notch patterns we are
+sure you will find deficiencies. When you do, github
+[issues](https://github.com/checkmate/simpl/issues/new) are accepted and
+[PR's](https://github.com/checkmate/simpl/pulls) are greatly appreciated.
+
+--------
+
+
+### The `simpl` vetting process
+
+`simpl` was conceived to solve a particular problem: that of duplicating code
+across multiple projects. As we moved from project to project we found certain
+patterns were being re-used by copying the code from one project to the next,
+which isn't too heinous until you reach about the third or fourth project at
+which point one begins to think of better ways to propagate useful patterns.
+Our solution was to create the `simpl` package: a place where we can collect
+these recurring patterns.
+
+Running with this idea, we have since learned that there are competing demands
+introduced by placing these patterns in a semi-official package:
+
+  - Because a pattern is being considered for inclusion in a package that is
+  intended to be used across many projects, a more rigorous review process
+  seems like a good idea.
+  - Identifying a pattern as one we have already implemented most often occurs
+  during a mad push to meet a deadline, so the person raising the PR requesting
+  the pattern be included in the `simpl` canon needs it merged quickly.
+
+__The following approach aims to reconcile these two competing demands:__
+
+  - The `simpl` package contains two types of modules:
+    - __Core__ modules are well-established, curated patterns that have been
+    thoroughly vetted. They have earned their place at the root level of the
+    `simpl` package. `config`, `exceptions`, `log` are all
+    examples of __Core__ modules.
+    - The `simpl.incubator` module contains modules that are being considered
+    for inclusion in __Core__, but are still being vetted. This solves the
+    copy/paste duplication issue while still allowing time for the rigor we
+    wish to apply to modules before being accepted into __Core__.
+  - Any submitted PR is a candidate for inclusion in __Core__.
+  - If a given PR is being contested, the submitter may opt to move the code
+  into `simpl.incubator` in an effort to keep the vetting process from blocking
+  usage of the code/pattern in other projects.
+  - `simpl.incubator` code can eventually be promoted to __Core__. If this
+  occurs the module will be moved to the root level of the `simpl` directory
+  and will no longer be available in `simpl.incubator`, requiring references to
+  the module to be updated when upgrading to the version of `simpl` that has
+  the module in __Core__.
+
+--------
+
+
+### Dealing with optional dependencies
+
+Some of `simpl`'s modules deal with specific tools (like MongoDB or Redis) and
+leverage Python packages built for those tools. This presents a problem,
+though: what if a project isn't using that particular `simpl` module? We don't
+want an unused module to force a project to install a package it has no
+intention of using!
+
+The following guidelines aim to solve this issue:
+  - When possible, limit package use to the
+  [Python Standard Library](https://docs.python.org/3/library/)
+  - When other packages are absolutely necessary make sure that the dependency
+  remains entirely encapsulated in the module that needs it. This means that
+  even though `simpl.bottle` requires the `bottle` package, neither a warning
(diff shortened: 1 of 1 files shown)


CONRTIBUTING.md
```diff
@@ -0,0 +1,9 @@
+# Guidelines for Contributing to `simpl`
+
+`simpl` was conceived to solve a particular problem: that of duplicating code
+across multiple projects. As we moved from project to project we found certain
+patterns were being re-used by copying the code from one project to the next,
+which isn't too heinous until you reach about the third or fourth project at
+which point one begins to think of better ways to propogate useful patterns.
```
samstav: propagate

samstav:
@Pablosan this is excellent.

CONRTIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+  - When other packages are absolutely necessary make sure that the dependency
+  remains entirely encapsulated in the module that needs it. This means that
+  even though `simpl.bottle` requires the `bottle` package, neither a warning
+  or exception should be thrown when importing `simpl.log` or `simpl.git` as
+  they can be used just fine without `bottle`.
+  - When other packages are optional (e.g. `simpl.middleware.cors` will use
+  `webob` but it isn't absolutely required), make sure that the module properly
+  handles both cases
+  ([cors.py](https://github.com/checkmate/simpl/blob/master/simpl/middleware/cors.py)
+  provides a good example of this).
```
samstav: Looking at cors.py, it actually seems like `webob

samstav:
What do we think about adopting angular's commit message guidelines?

https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit

If we like this, CONTRIBUTING.md would be a good place to mention it.

CONRTIBUTING.md
ziadsawalha: The example we gave is that you could `import simpl.cors` and then either read `CORSMiddleware.default_headers` or subclass it and override the `__call__` method. Not the most visibly useful of examples, but somewhat illustrative of the concept.

Is there a better example already in simpl that you have in mind?

larsbutler:
+1 to commit message guidelines. I'm all for documenting guidelines as much as possible, so that there are fewer issues to discuss in a PR. In other words, instead of debating over issues like style (code and commit message), code reviewers are simply there to ensure that the guidelines are enforced--but the guideline, the document, is the source of truth.

ziadsawalha:
+1 to that. We used them in [chessboard](https://github.com/checkmate/chessboard/blob/master/CONTRIBUTING.md) - That file has been trimmed down to match some of what we already do so we could use it as a starting point.

larsbutler:
@Pablosan What about test coverage guidelines? I know it's not always possible or feasible to get 100%, but I think we should shoot for a high number and try to enforce that as much as possible. I personally feel that testing should be as much of a part of the style guide as PEP8 is.

CONRTIBUTING.md
samstav: Ah, I see. Makes perfect sense. (you might have to s-p-e-l-l it out for kids like me). Thus example is good then. What about keyring in config.py ?

larsbutler:
Another thing I would like to see: Some insight into our design and decision making process. We decided that we want to do "thoughtful, inclusive, and deliberate architecture/UI/UX/Product design". I think that has a place here as well.

For a relevant reference, check out the [C4 contribution guidelines](http://rfc.zeromq.org/spec:22) for ZeroMQ. I'm not suggesting we follow all of this, but I think there are some good things we can take from it. This is one of my favorites (from the `Patch Requirements` section):

`A patch SHOULD be a minimal and accurate answer to exactly one identified and agreed problem.`

Thoughts?

ziadsawalha:
@larsbutler I think that applies to and caters for https://github.com/checkmate/simpl/pull/47/files#diff-c7ea04263ae1b15a0baf530bce5039ebR33. But https://github.com/checkmate/simpl/pull/47/files#diff-c7ea04263ae1b15a0baf530bce5039ebR36 is a reality we need to also cater for.

So we're suggesting that the criteria for accepting something into `simpl.incubator` be:
- it exists in more than one of our repos already (or it exists in one of our repos already and is about to be used in another).
- the PR includes the aggregated code and tests from of the copies in the other repos (and that it can be used instead of the copies).

And the reviews should focus only tests, bugs, python/style tweaks, but above all not stall on design/pattern discussions.

Pablosan:
I really like the ideas I'm seeing for broadening the scope of topics
covered in this document. The version I have proposed is by no means
comprehensive.

I propose that we work toward consensus on what is there now, because this
will help solve an immediate problem. Then any of us should feel free to
propose additions and improvements in subsequent PR's.

I'm concerned that if we attempt to be comprehensive up front that this
will become a very long process. My preference is to nail down what's there
now and iterate.

graywizardx:
I like the CONTRIBUTING.md I think thats the right approach, but if I might play devils advocate for a moment: Why are we not dividing this library up so that it always has hard dependencies? It seems we are doing extra work by trying to carefully isolate dependency chains when we could just have multiple libraries. 

Sam brought up previously the cost of multiple pypi libraries/projects but in this case my gut says that it will make our lives better in the long run.

ziadsawalha:
@graywizardx wrt https://github.com/checkmate/simpl/pull/47#issuecomment-118069097: can you elaborate on "it will make our lives better in the long run"?

graywizardx:
In looking over the Simpl code it appears to me that it is a number of specific purpose libraries (such as config handling) that are being rolled into a single library for convenience. In general its probably not a bad idea, but it seems we are going through machinations to prevent dependency issues, as well as looking at procedural structure to keep the intention of the library pure, even with multiple possibly competing requirements (here in the form of library dependencies).

I think if we split the library into multiple independent libraries (Config, Server Framework, etc) those independent libraries could make whatever demands on code and dependencies they felt where required and the scope of impact would be that particular library only.

As these libraries grow and mature we would not need to worry as much about making sure that library requirements didnt overlap or compete or about how to balance conflicting priorities in regards to structure or dependencies.

Pablosan:
When we first talked about starting `simpl` I had envisioned a library that only had dependencies on the Python Standard Library. We talked about this as a team at length some time ago and I backed off on that constraint and agreed with the team that, as long as we didn't force package installs that were only needed for certain modules that may or may not be used in a given situation, I was fine with optional packages.

I agree that this does create overhead in `simpl`, but it doesn't look to me like the overhead is really that much. I tend to prefer encapsulation so I am not opposed to separate repos for things, but multiple repos comes with overhead too.

My current view of `simpl` is that it provides a way for us to collect repeating patterns for easy re-use, and that those patterns are likely either not quite ready to be their own package or have already been codified well somewhere else. By collecting these patterns in one place, my hope is that it makes it easier to identify those patterns which are robust and mature enough to warrant promotion to their own repo, and to identify those patterns which we would do well to replace with top-notch external packages.

graywizardx:
I wouldnt take my words with too much emphasis, I am just a bit weary of libraries full of helper functions. Amz had the propensity to build "wrappers" for common functionality because someone thought they needed something custom, or a team/org decided that functionality needed to work a certain way. Over the years these collapsed into a few "key" libraries that were either better served by newer more modular libraries or by now built in language features.

I too am a big fan of building libraries with minimal dependencies but I also want libraries that are force multipliers as I work. A library that brings in its own dependencies is *ok* with me, but I generally prefer those dependencies will not conflict with what I am actually doing. For instance if you bring in an AST helper and Im building a webserver, no issue. If you bring in Scipy and I am writing a data cruncher we might have issues. 

I guess what I am really driving at is I would hope for Simpl to be things that accelerated my workflow and I was not worried about them interfering with my work. If I was pulling them in to gain abilities that were difficult or cumbersome to reimplement then I get a lot of perceived value. If the library is a common pattern repository then I would push for it to be specific libraries that could be imported by need.

PR #52, opened 2015-07-03
feat(server): add bottle server module

ziadsawalha:
Includes code for eventlet and tornado adapters
with customizations.

xtornado: enables xheaders always.
xeventlet: supports SSL for testing and additional
    tuning parameters.

### DIFF
--- a/all-requirements.txt
+++ b/all-requirements.txt
@@ -3,5 +3,9 @@
 # For simpl.rest
 bottle==0.12.8
 
+# For simpl.server
+eventlet==0.17.4
+tornado==4.2
+
 # For simpl.middleware.cors
 WebOb==1.4.1
--- a/simpl/server.py
+++ b/simpl/server.py
@@ -1,0 +1,171 @@
+# Copyright 2013-2015 Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Bottle Server Module."""
+
+import logging
+import os
+
+import bottle
+
+LOG = logging.getLogger(__name__)
+
+
+class EventletLogFilter(object):  # pylint: disable=R0903
+
+    """Receives eventlet log.write() calls and routes them.
+
+    Thread and WSGI messages only get written in debug mode. Eventlet writes
+    these out starting with a thread id in parens or "wsgi" messages:
+    - "(46722) wsgi starting up on http://127.0.0.1:8080/"
+    - "(47001) accepted ('127.0.0.1', 53046)"
+    - "wsgi exiting"
+
+    All other calls are assumed to be apache log-style API calls and we want
+    these written as INFO and to an access_log if provided.
+
+    API calls are assumed to start with the IP address:
+    - "127.0.0.1 - - [07/Jul/2015 16:16:31] "GET /version HTTP/1.1" 200 ..."
+
+    An instance of this class can be passed in to a bottle.run command using
+    the `log` keyword.
+    """
+
+    def __init__(self, log, access_log=None):
+        """Initialize with config and optional access_log.
+
+        :param log: logger instance (ex. logging.getLogger()).
+        :keyword access_log: a file handle to an access log that should receive
+            apache-style entries for each call and response.
+        """
+        self.log = log
+        self.access_log = access_log
+
+    def write(self, text):
+        """Write to appropriate target."""
+        if text:
+            if text[0] in '(w':
+                # write thread and wsgi messages to debug only
+                self.log.debug(text[:-1])
+                return
+            if self.access_log:
+                self.access_log.write(text)
+            self.log.info(text[:-1])
+
+
+class XEventletServer(bottle.ServerAdapter):
+
+    r"""Eventlet Bottle Server Adapter with extensions.
+
+    Supports SSL. Accepts additional tuning parameters:
+
+    * `backlog` adjust the eventlet backlog parameter which is the maximum
+      number of queued connections. Should be at least 1; the maximum
+      value is system-dependent.
+    * `family`: (default is 2) socket family, optional. See socket
+      documentation for available families.
+    * `**kwargs`: directly map to python's ssl.wrap_socket arguments from
+      https://docs.python.org/2/library/ssl.html#ssl.wrap_socket and
+      wsgi.server arguments from
+      http://eventlet.net/doc/modules/wsgi.html#wsgi-wsgi-server
+
+    To create a self-signed key and start the eventlet server using SSL::
+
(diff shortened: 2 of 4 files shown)


simpl/server.py
```diff
@@ -0,0 +1,158 @@
...
+
+import logging
+import os
+
+import bottle
+
+LOG = logging.getLogger(__name__)
+
+
+class EventletLogFilter(object):  # pylint: disable=R0903
```
larsbutler: Why do we have `pylint: disable` statements in `simpl`? We're not even using `pylint`, as far as I can tell (despite the fact that we have a pylintrc file).

ziadsawalha: I prefer to be explicit. Especially when doing anything that can be flagged by one of the common linters.

larsbutler: But it doesn't actually do anything. Why don't start using pylint then? Otherwise, we're just adding superfluous comments to the codebase.

simpl/server.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, conf, log, access_log=None):
+        """Initialize with config and optional access_log."""
+        self.config = conf
+        self.log = log
+        self.access_log = access_log
+
+    def write(self, text):
+        """Write to appropriate target."""
+        if text:
+            if text[0] in '(w':
```
larsbutler: What is the significance of `(w`?

larsbutler:
`server.py` has 23% test coverage.

```
simpl.config              347     43    88%   259, 264-265, 330-333, 339, 376-378, 390, 396, 425, 429, 433, 437, 444, 487, 510-511, 568, 593-597, 632, 636-637, 648, 661, 811-816, 835-842, 860-864, 867
simpl.git                 299     10    97%   139, 199-200, 281, 333, 431, 541, 729-730, 732
simpl.incubator             1      0   100%   
simpl.incubator.dicts     147      7    95%   49, 68, 147-148, 155, 248, 261
simpl.log                  49     32    35%   87-94, 104-111, 124-133, 140-153, 165, 169, 174, 177
simpl.rest                 82      3    96%   39, 54, 192
simpl.secrets              12      5    58%   32-36
simpl.server               70     54    23%   38-40, 44-51, 83-109, 113-135, 139, 150-156
TOTAL                    1162    168    86%   

```

simpl/server.py
```diff
@@ -0,0 +1,158 @@
...
+                    'ssl_version', 'ca_certs', 'do_handshake_on_connect',
+                    'suppress_ragged_eofs', 'ciphers'):
+            try:
+                ssl_args[arg] = self.options.pop(arg)
+            except KeyError:
+                pass
+        address = (self.host, self.port)
+        try:
+            sock = eventlet.listen(address, **socket_args)
+        except TypeError:
```
larsbutler: What causes this to throw a `TypeError`?

simpl/server.py
ziadsawalha: You raise a valid question which we should bring up for the team to discuss.  Most of our other projects run pylint in the tox runs. I would vote to include it (with a customized pylintrc file). I've raised #55 to host that discussion.

WRT to this PR, pylint comments are in many other places in this codebase (as well as in many of our other projects - most of which actually include pylint in the tox runs) - as well as a pylintrc file - so this line does not diverge from what we've been doing.

It does document that we've intentionally not included public methods in this class. And it does  it in a way that is compatible with one of the most commonly used linting tools for python.

While removing this line is simple, I'd rather stay consistent with the rest of the codebase and remove it if we choose to not use pylint and remove pylint comments from the whole code base.

simpl/server.py
ziadsawalha: It's what eventlet preceds its WSGI event messages with; we write those as debug messages. The other messages are apache log-style API messages.

ziadsawalha: This is ported from https://github.com/bottlepy/bottle/blob/master/bottle.py#L2997. Based on the comment I would say some version of eventlet.

simpl/server.py
ziadsawalha: This is ported from https://github.com/bottlepy/bottle/blob/master/bottle.py#L2997. Based on the comment I would say some version of eventlet.

ziadsawalha:
@larsbutler WRT test coverage, I started down the route of copying over tests from https://github.com/bottlepy/bottle. Before I go down that route, is that an acceptable approach?

FWIW, it calls out to subprocesses to accomplish the testing so it won't impact coverage stats. I don't know if test coverage is the metric we want to strive for in this case.

simpl/server.py
larsbutler: Ah, I never would have guessed that. Is this documented somewhere?

simpl/server.py
```diff
@@ -0,0 +1,158 @@
...
+    """Receives eventlet log.write() calls and routes them.
+
+    Thread and WSGI messages only get written in debug mode.
+    API calls get written as INFO and to an access_log if provided.
+
+    An instance of this class can be passed in to a bottle.run command using
+    the `log` keyword.
+    """
+
+    def __init__(self, conf, log, access_log=None):
```
larsbutler: Please add `param/keyword` docs for these parameters. I could make a guess at what `log` is, but it's not clear what the others are.

Also, `conf/self.config` doesn't appear to be used anywhere.

simpl/server.py
```diff
@@ -0,0 +1,158 @@
...
+        self.config = conf
+        self.log = log
+        self.access_log = access_log
+
+    def write(self, text):
+        """Write to appropriate target."""
+        if text:
+            if text[0] in '(w':
+                # write thread and wsgi messages to debug only
+                self.log.debug(text[:-1])
```
larsbutler: What are we chopping off here? What's is the expected value of the last item?

simpl/server.py
```diff
@@ -0,0 +1,158 @@
...
+    the `log` keyword.
+    """
+
+    def __init__(self, conf, log, access_log=None):
+        """Initialize with config and optional access_log."""
+        self.config = conf
+        self.log = log
+        self.access_log = access_log
+
+    def write(self, text):
```
larsbutler: What is `text` exactly? I would expect it to be a string, but below it's treated like a list. There's also no test code which exercises it, so I have no way of knowing a) what the method expects; b) how I'm supposed to use the class; or c) whether or not it is implemented correctly.

simpl/server.py
ziadsawalha: Added docs and tests

simpl/server.py
ziadsawalha: No, but I added the docs now.

simpl/server.py
ziadsawalha: We're chopping off the carriage return

all-requirements.txt
```diff
@@ -3,5 +3,9 @@
 # For simpl.rest
 bottle==0.12.8
 
+# For simpl.server
+eventlet==0.17.4
+tornado==4.2
```
Pablosan: I know the pattern predates this PR, but would it make more sense for module-specific requirements to be listed in a module-specific requirements file?

ziadsawalha: May I submit another PR for that after this?

ziadsawalha:
@larsbutler I added integration tests for the Tornado and Eventlet server adapters. Because they fork, they don't show up as covered. I also added tests for the Eventlet log writer.

all-requirements.txt
Pablosan: Yes, absolutely. I'm not even sure separate files is the right approach. We could use a follow-up PR to facilitate that discussion.

PR #55, opened 2015-07-06
chore(style): add pylint checks

ziadsawalha:
We use pylint in most of our codebases, many devs
add it to their dev cycle, and we commonly include pylint comments in our code which can help declare
intent when we have code that could be flagged by
pylint.

Notes: this does not block builds or tests

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,7 +1,8 @@
 [Messages Control]
-# W0511: TODOs in code comments are fine.
-# W0142: *args and **kwargs are fine.
-disable=W0511,W0142
+enable=C,E,F
+disable=W,R,I,fixme,too-many-arguments,wildcard-import,no-absolute-import,bad-builtin,import-error
+# NOTE(larsbutler): import-error is a spurious message from a bug, it seems.
+# See https://bitbucket.org/logilab/pylint/issues/324/namespaces-and-virutlaenv.
 
 [BASIC]
 # Don't require docstrings on tests or Python Special Methods
--- a/simpl/git.py
+++ b/simpl/git.py
@@ -600,6 +600,7 @@ class GitRepo(object):
         """Create an annotated tag."""
         return git_tag(self.repo_dir, tagname, message=message, force=force)
 
+    # pylint: disable=invalid-name
     def ls(self):
         """Return a list of *all* files & dirs in the repo.
 
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -6,5 +6,6 @@ flake8_docstrings
 mock
 nose
 nose-ignore-docstring
+pylint
 requests
 webtest
--- a/tox.ini
+++ b/tox.ini
@@ -18,8 +18,7 @@ commands = nosetests {posargs} --with-doctest --with-coverage --cover-package=si
 # - tests may not have docstrings
 commands =
     flake8 setup.py simpl
-    # Skip linting errors; just write out a file that can be used for analysis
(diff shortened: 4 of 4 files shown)


larsbutler:
I think we should either gate on pylint (and get a good, well-tuned pylint config) or not use it at all. I'm not a big fan of the "in-between" solution.

ziadsawalha:
You also don't want to end up with code full of `pylint: disable` comments that are there just to pass the gate. "well-tuned" is key. We've historically gated on the total number of pylint W/E/R errors, but not forced it to be zero. Hence the `|| true` in tox.ini. That requires code or a tool (like Jenkins) than can parse, track, report, and gate on pylint error count.

larsbutler:
I did some experimentation with this branch. It seems like the main checks we want pylint to do are C, E, and F (Convention, Error, and Fatal, respectively). There are a few exceptions. The R (Refactor) warnings are ostensibly useful, but they're mostly just "too-many-X" warnings, which I find to be a bit silly and arbitrary.

By playing around with the config and tweaking a couple of things, I came up with a small patch that gives us a perfect score, but keeps many of the critical checks in place:

```diff
diff --git a/pylintrc b/pylintrc
index 192b684..bf63a0c 100644
--- a/pylintrc
+++ b/pylintrc
@@ -1,7 +1,8 @@
 [Messages Control]
-# W0511: TODOs in code comments are fine.
-# W0142: *args and **kwargs are fine.
-disable=W0511,W0142
+enable=C,E,F
+disable=W,R,I,fixme,too-many-arguments,wildcard-import,no-absolute-import,bad-builtin,import-error
+# NOTE(larsbutler): import-error is a spurious message from a bug, it seems.
+# See https://bitbucket.org/logilab/pylint/issues/324/namespaces-and-virutlaenv.
 
 [BASIC]
 # Don't require docstrings on tests or Python Special Methods
diff --git a/simpl/git.py b/simpl/git.py
index 8d401a1..e501d77 100644
--- a/simpl/git.py
+++ b/simpl/git.py
@@ -600,6 +600,7 @@ class GitRepo(object):
         """Create an annotated tag."""
         return git_tag(self.repo_dir, tagname, message=message, force=force)
 
+    # pylint: disable=invalid-name
     def ls(self):
         """Return a list of *all* files & dirs in the repo.
 
diff --git a/tox.ini b/tox.ini
index 2df5a61..3392994 100644
--- a/tox.ini
+++ b/tox.ini
@@ -18,8 +18,7 @@ commands = nosetests {posargs} --with-doctest --with-coverage --cover-package=si
 # - tests may not have docstrings
 commands =
     flake8 setup.py simpl
-    # Skip linting errors; just write out a file that can be used for analysis
-    /bin/bash -c "pylint -f parseable {posargs:simpl} > pylint.log || true"
+    pylint simpl
 
 [flake8]
 #flake8 default settings
```

Let me know what you think!

samstav:
@larsbutler this rc would disable all warning, refactor, and info messages?

larsbutler:
@samstav Correct.

larsbutler:
@samstav Those are the messages which are, in my opinion, not so important. flake8 should catch a lot of that.

ziadsawalha:
@larsbutler - that's great! Thanks. I'm all for get something in and then tweak as we see fit. I'll update the patch.

larsbutler:
LGTM, although I would like to know what @Pablosan thinks of this configuration, since he has opinions about pylint.

Pablosan:
@larsbutler @ziadsawalha this looks great to me! Thanks for doing the research on this Lars. It looks very balanced to me.

PR #61, opened 2015-07-09
a copyable Config object for python 3

### LINKED ISSUE #60: Config objects are not copyable in python 3 due to infinite recursion
```python
import copy

from simpl import config

copy.copy(config.Config())
```
throws the following exception:

```pytb
/Users/samstav/.virtualenvs/py3-dev/lib/python3.4/copy.py in copy(x)
    103                 raise Error("un(shallow)copyable object of type %s" % cls)
    104
--> 105     return _reconstruct(x, rv, 0)
    106
    107

/Users/samstav/.virtualenvs/chefproxy-dev-py3/lib/python3.4/copy.py in _reconstruct(x, info, deep, memo)
    299         if deep:
    300             state = deepcopy(state, memo)
--> 301         if hasattr(y, '__setstate__'):
    302             y.__setstate__(state)
    303         else:

/Users/samstav/.virtualenvs/py3-venv/lib/python3.4/site-packages/simpl/config.py in __getattr__(self, attr)
    439     def __getattr__(self, attr):
    440         """Get attribute."""
--> 441         if attr in self._values:
    442             return self._values[attr]
    443         else:

...

/Users/samstav/.virtualenvs/py3-venv/lib/python3.4/site-packages/simpl/config.py in __getattr__(self, attr)
    439     def __getattr__(self, attr):
    440         """Get attribute."""
--> 441         if attr in self._values:
    442             return self._values[attr]
    443         else:

/Users/samstav/.virtualenvs/py3-venv/lib/python3.4/site-packages/simpl/config.py in __getattr__(self, attr)
    439     def __getattr__(self, attr):
    440         """Get attribute."""
--> 441         if attr in self._values:
    442             return self._values[attr]
    443         else:

RuntimeError: maximum recursion depth exceeded while calling a Python object
```

I dug around for a bit between the implementations of copy.py, but I couldn't quite nail it down.

http://nedbatchelder.com/blog/201010/surprising_getattr_recursion.html

samstav:
Fixes #60

### DIFF
--- a/simpl/config.py
+++ b/simpl/config.py
@@ -438,6 +438,10 @@ class Config(collections.MutableMapping):
 
     def __getattr__(self, attr):
         """Get attribute."""
+        # protection from infinite recursion when __init__
+        # is skipped during object creation (probably copy.copy)
+        if attr == '_values':
+            raise AttributeError()
         if attr in self._values:
             return self._values[attr]
         else:
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -18,6 +18,7 @@
 from __future__ import print_function
 
 import argparse
+import copy
 import errno
 import os
 import string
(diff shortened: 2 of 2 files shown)


Pablosan:
:+1:  LGTM

simpl/config.py
```diff
@@ -438,6 +438,8 @@ def __len__(self):
 
     def __getattr__(self, attr):
         """Get attribute."""
+        if attr == '_values':
+            raise AttributeError()
```
ziadsawalha: Wow. This was an intricate bug. Is it worth documenting why we're doing this?

```
        # avoid infinite recursion if ._values doesn't exist which can happen when __init__
        # is skipped during deserialization or certain types of copy operations
        # Ref: the URL
```

Or in the commit message.

samstav: I'll add a comment and to the commit message. Thanks

PR #62, opened 2015-07-09
Add coveralls support

Gifflen:
Since we are already outputting coverage data in tox I've added support for historically tracking coverage with coveralls.io. There are also badges for travis and coveralls on the README.md


samstav:
Sweeeet.

PR #63, opened 2015-07-10
chore(dicts): simplify list generation code

mgeisler:
The old code would build a range (a list) and iterate over it, binding
an unused name in the process.

The new code multiplies a singleton list, which is handled internally
as a simple for-loop that copies the single element into all entries
in the new list.


mgeisler:
I'm unsure how the test failure here can be caused by my change...

larsbutler:
Yeah, that failure doesn't seem to be related at all. Odd.

ziadsawalha:
Issues fixed in PR #64

ziadsawalha:
Thank you, @mgeisler

PR #66, opened 2015-07-14
[incubator.rest] Request schema validation deco 

larsbutler:
It's a work-in-progress.

How to use:

```python
import bottle

from simpl import rest

import voluptuous as volup


FOO_QUERY_SCHEMA = volup.Schema({
    'a': rest.coerce_many(int),
    'b': rest.coerce_one(str),
})


@bottle.route('/foo')
@rest.schema(query_schema=FOO_QUERY_SCHEMA)
def foo(body=None, query=None):
    return dict(body=body, query=query)


def start():
    """Start a Juno Server."""
    app = bottle.default_app()
    bottle.run(app=app, host='0.0.0.0', port=8080)

if __name__ == '__main__':
    start()
```

Since some endpoints might want to accept multiple values for a given query parameter, the `One` and `Many` functions allow that to be specified and handled nicely.

This is somewhat opinionated for `bottle`, since query params from bottle requests are group into dicts of lists, keyed by the parameter name.

### DIFF
--- a/all-requirements.txt
+++ b/all-requirements.txt
@@ -2,6 +2,7 @@
 
 # For simpl.rest
 bottle==0.12.8
+voluptuous==0.8.7
 
 # For simpl.server
 eventlet==0.17.4
--- a/simpl/incubator/rest.py
+++ b/simpl/incubator/rest.py
@@ -1,0 +1,192 @@
+# Copyright (c) 2011-2015 Rackspace US, Inc.
+# All Rights Reserved.
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+"""Incubator utilities for :mod:`simpl.rest`."""
+
+import bottle
+import six
+import voluptuous as volup
+
+
+class MultiValidationError(Exception):
+
+    """Basically a re-imagining of a `voluptuous.MultipleInvalid` error.
+
+    Reformats multiple errors messages for easy debugging of invalid
+    Checkmatefiles.
+    """
+
+    def __init__(self, errors):
+        """MultiValidationError constructor.
+
+        :param errors:
+            List of `voluptuous.Invalid` or `voluptuous.MultipleInvalid`
+            exception objects.
+        """
+        self.errors = errors
+        self.message = self._generate_message()
+
+    def __str__(self):
+        """Just return the pre-computed message.
+
+        See :meth:`_generate_message`.
+        """
+        return self.message
+
+    def __repr__(self):
+        """Simple representation of the exception, with the full message."""
+        indented_message = '\n'.join(
+            sorted('\t' + x for x in self.message.split('\n'))
+        )
+        return (
+            '%(cls_name)s(\n%(message)s\n)'
+            % dict(cls_name=self.__class__.__name__, message=indented_message)
+        )
+
+    def _generate_message(self):
+        """Reformat `path` attributes of each `error` and create a new message.
+
+        Join `path` attributes together in a more readable way, to enable easy
+        debugging of an invalid Checkmatefile.
+
+        :returns:
+            Reformatted error paths and messages, as a multi-line string.
+        """
+        reformatted_paths = (
+            ''.join(
+                "[%s]" % str(x)
+                # If it's not a string, don't put quotes around it. We do this,
+                # for example, when the value is an int, in the case of a list
+                # index.
+                if isinstance(x, six.integer_types)
+                # Otherwise, assume the path node is a string and put quotes
+                # around the key name, as if we were drilling down into a
+                # nested dict.
+                else "['%s']" % str(x)
+                for x in error.path
+            )
+            for error in self.errors
+        )
+
+        messages = (error.msg for error in self.errors)
+        # combine each path with its message:
+        zipped = zip(reformatted_paths, messages)
+        combined_messages = (
+            '%(path)s: %(messages)s' % dict(path=path, messages=message)
+            for path, message in zipped
+        )
+
+        return '\n'.join(sorted(combined_messages))
+
+
+def coerce_one(schema=str):
+    """Expect the input sequence to contain a single value.
+
+    :keyword schema:
+        Custom schema to apply to the input value. Defaults to just string,
+        since this is designed for query params.
+    """
+    def validate(val):
+        """Unpack a single item from the inputs sequence and run validation.
+
+        NOTE(larsbutler): This code is highly opinionated for bottle, since
+        bottle query params are wrapped in a list, even if there is just a
+        single value for a given parameter.
+        """
+        [value] = val
+        return volup.Coerce(schema)(value)
+    return validate
+
+
+def coerce_many(schema=str):
+    """Expect the input to be a sequence of items which conform to `schema`."""
+    def validate(val):
+        """Apply schema check/version to each item."""
+        return [volup.Coerce(schema)(x) for x in val]
+    return validate
+
+
+def schema(body_schema=None, body_required=False, query_schema=None,
+           content_types=None, default_body=None):
+    """Decorator to parse and validate API body and query string.
+
+    This decorator allows one to define the entire 'schema' for an API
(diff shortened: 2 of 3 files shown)


larsbutler:
This is more or less copies the `body` decorator and extends it to include query schema.

ziadsawalha:
Why not also pass the query params in as kwargs since you already have them defined?

```python
@rest.schema(query_schema=FOO_QUERY_SCHEMA)
def foo(body, a=None, b=None):
    return body
```

larsbutler:
@ziadsawalha Yes, you're right, that's what should be done.

larsbutler:
@ziadsawalha Actually, the deco passes the entire query dict as a single arg. I'm not sure if it should be an arg or a kwarg.

larsbutler:
I'm thinking kwargs would be best for both the body and query; that way they can be optional. Here we go:

```python
import bottle

from simpl import rest

import voluptuous as volup


FOO_QUERY_SCHEMA = volup.Schema({
    'a': rest.Many(int),
    'b': rest.One(str),
})


@bottle.route('/foo')
@rest.schema(query_schema=FOO_QUERY_SCHEMA)
def foo(body=None, query=None):
    return body


def start():
    """Start a Juno Server."""
    app = bottle.default_app()
    bottle.run(app=app, host='0.0.0.0', port=8080)

if __name__ == '__main__':
    start()
```

larsbutler:
In case it wasn't clear, I've still got a lot of docs and tests to add, so please don't merge anything yet.

simpl/rest.py
```diff
@@ -23,6 +23,77 @@
...
+                body = default_response
+            if body_schema:
+                try:
+                    body = body_schema(body)
+                except Exception as exc:
+                    bottle.abort(400, str(exc))
+
+            # validate the query string per the schema (if application):
+            query = bottle.request.query
+            validated_query = query_schema(query.dict)
```
ziadsawalha: This will return a 500. We can catch an error and return a 400 with the useful info. For example, in the body call, you get an error like `extra keys not allowed @ data['foo']`, especially if the callable is a voluptuous schema (catch MultipleInvalid). But just in case it is not, we should also look into catching Value, Type, and IndexErrors.

simpl/rest.py
```diff
@@ -23,6 +23,77 @@
 STANDARD_QUERY_PARAMS = ('offset', 'limit', 'sort', 'q', 'facets')
 
 
+def One(schema=str):
```
ziadsawalha: Did you know about [voluptuous.Length](https://github.com/alecthomas/voluptuous/blob/master/voluptuous.py#L1502)? It might return a more useful error.

simpl/rest.py
```diff
@@ -23,6 +23,77 @@
...
+    """
+
+    def f(v):
+        """"""
+        [value] = v
+        return schema(value)
+    return f
+
+
+def Many(schema=str):
```
ziadsawalha: What does this do?

simpl/rest.py
ziadsawalha: How can we integrate this with the pagination code?

samstav:
@larsbutler: IIRC, there was some work you did (I believe it was in chessboard or checkmate) to wrap validation errors from voluptuous to make them more useful? Is this something you'd like to include here?

larsbutler:
@samstav Sure, I could include that.

simpl/rest.py
larsbutler: The purpose of `Many` is to specify a sequence of values for bottle query params. After looking at this again, it could be replaced with `[str]` or `[sometype]`.

larsbutler: I'll double check that.

larsbutler: Nope, it doesn't work the same way. I'm working on tests right now, which should make the utility of `Many` a bit clearer.

ziadsawalha:
Sorry for bringing this up after you've run some cycles on it, but I was just using [`process_params`](https://github.com/checkmate/simpl/blob/master/simpl/rest.py#L235) and realized it has much of the logic that we've iterated on for handling pagination and multiple option params. So that got me wondering if much of this functionality isn't already there between `process_params` and `body`and that maybe we just need to enhance or integrate them instead of creating a new decorator? Or at least we should document how the three of them are used together to make sure they are complementary, comprehensive, and somewhat intuitive.

larsbutler:
@ziadsawalha Hmmm, I'll see what can be done.

larsbutler:
Well, the one thing I can say about `process_params` is that bottle does some of this already, by grouping parameters together. It also doesn't seem to have the capability to enforce parameter schema. For example, some filter params may appear multiple times, while other params should appear only once.

What do you want to do about that?

ziadsawalha:
Our APIs commonly support the following ways to filter on multiple values (ex. statuses):
- one per param: `?status=OK&status=WARNING`
- comma-separated: `?status=OK,WARNING`
- both simultaneously: `?status=OK,WARNING&status=ERROR`

And we support pagination with `limit` and `offset`.

Much of the logic in `process_params` validates and normalizes these formats into a simple dict (ex. any fields marked as `filter_fields` get processed and passed to the route as a simple, normalized list).

I don't believe bottle does that.

Off the top of my head, here are some possible ways we can address this:
1 - call `process_params` first inside the decorator and pass the dict it returns to `query_schema`. This requires adding the process_params signature to the decorator... which might get messy.
2 - incorporate some of the logic in `process_params` into the decorator code. This might still need more args/kwargs than we'd like in the decorator.
3 - re-implement some of the logic in `process_params1 using custom schema elements and use those in the decorator. For example, here's a schema mock-up:
```python
QUERY_SCHEMA = {
    'status': rest.MultiFilter(max=4),  # <-- custom filter that handles multiple and/or comma-separated values
    'limit': volup.Range(min=0, max=1000),
    'offset': volup.Range(min=0, max=1000000),
    ...
}
```
I think these happen to be sorted in ascending order of difficulty (amount of work to get them done), but I also believe 3 will result in the cleanest interface (similar to the one you have already).

larsbutler:
So, what would be the main thing that `schema` needs to add? The logic to split groups of parameters on a comma? Is that it, or is there something more specific?

simpl/rest.py
```diff
@@ -41,6 +43,179 @@ def __init__(self, message, http_code=400, reason=None):
...
+                if query_schema is not None:
+                    try:
+                        query = query_schema(query)
+                    except volup.MultipleInvalid as exc:
+                        raise MultiValidationError(exc.errors)
+                if not query:
+                    # If the query dict is empty, just set it to None.
+                    query = None
+
+                # Pass `body` and `auery` as kwargs to the decorated function.
```
ziadsawalha: typo: `auery`

tests/test_rest.py
```diff
@@ -443,5 +445,168 @@ def test_yaml(self):
...
+
+        def callback(body=None, query=None):
+            return dict(body=body, query=query)
+        self.callback = callback
+
+    def test_no_schema(self):
+        # Use the decorator without the @ syntax:
+        self.callback = rest.schema()(self.callback)
+
+        self.root_app.route('/foo', callback=self.callback)
```
ziadsawalha: Does this route get overwritten in the same app for every test since setup uses the `default_app()`? Maybe use `self.root_app = bottle.Bottle()` to create a new instance for each tests is "safer"? I'm not seeing issues, but it _might_ break intermittently or hide problems if/when any tests run concurrently or some garbage is left over in the app...

larsbutler: Oh, is `default_app` a singleton?

samstav: Yes. It is far from least astonishment. I'm not a fan.

ziadsawalha:
The uses we designed it to handle are included in the [tests](https://github.com/checkmate/simpl/blob/master/tests/test_rest.py#L243). Apart from the fields above, there's some support for faceting (code pending porting to simpl), sorting, text searching params and defaults.

tests/test_rest.py
ziadsawalha: I understand why it is there, though. It allows any module to add a route by just decorating a function with `@get`.

tests/test_rest.py
```diff
@@ -19,6 +19,8 @@
 import bottle
 import mock
 import six
+import voluptuous as volup
+import webtest
```
ziadsawalha: webtest is already imported below (in the third party imports section)

tests/test_rest.py
```diff
@@ -19,6 +19,8 @@
 import bottle
 import mock
 import six
+import voluptuous as volup
```
ziadsawalha: This is not used.

larsbutler: Ah, that's leftover from a the rebase. I'll clean these up.

tests/test_rest.py
larsbutler: `webtest` should be in the middle section. But there's no reason to touch this file, so I'll revert it.

PR #69, opened 2015-07-20
(WIP) Query and body decorators

larsbutler:


samstav:
Will you be removing [this code](https://github.com/checkmate/simpl/blob/1a3c1dde8e6a18dd785f1711ddb0af8e23bf2341/simpl/rest.py#L44-L79) in this PR?

ziadsawalha:
@samstav - on the contrary; we're removing the duplicate logic from the PR. So the `query` decorator handles query params (applies to all methods) and the `body` decorator will be used for methods that have a body (not GET, OPTIONS, etc...). They become complementary decorators.

samstav:
@ziadsawalha So I can ignore the definition (additions) of the `body()` function in this PR?

ziadsawalha:
Oh, I see what you mean @samstav - I assume that's a copy/paste error.

ziadsawalha:
Actually, @samstav, looking at the commit message:
> The only major change here is that the decorator injects the `body` as a
kwarg, instead of a positional arg.

simpl/incubator/rest.py
```diff
@@ -190,3 +154,41 @@ def wrapped(*args, **kwargs):
...
+            if body is None:
+                body = default
+            if schema:
+                try:
+                    body = schema(body)
+                except (KeyboardInterrupt, SystemExit):
+                    raise  # don't catch and ignore attempts to end the app
+                except Exception as exc:
+                    bottle.abort(400, str(exc))
+            return fxn(*args, body=body, **kwargs)
```
ziadsawalha: Why do we want to change this to a kwarg?

larsbutler: This seemed like a better approach, since we're injecting several parameters into a function. With kwargs, the order doesn't matter. With posargs, the order does matter, so injecting these are kwargs seem a little less invasive/error prone.

ziadsawalha: I understand "less invasive" as being less restrictive on the decorated callable's signature. I agree that adding a kwarg (as opposed to adding an arg) is easier to do, easier to move around or handle using `**kwargs`, and can be done with more tolerance for argument ordering (to some degree, since kwargs can still receive values sent in as positional argument). But I do think it does also have the downside that it can conflict with kwargs or args of the same name if they are already in the decorated callable signature causing `multiple values for keyword argument 'body'` errors. And while that can be partially resolved by checking for duplicate kwargs like so:
```python
def decorator(fxn):
    def wrapper(*args, **kwargs):
        kwargs.setdefault('body', new_data)  # or `kwargs['body'] = new_data
        return fxn(*args, **kwargs)
    return wrapper
```
there are cons to either of those fixes that could cause hard-to-find "_where is the data I am passing in the `body` argument getting changed?!_" type of bugs.       

If this was a "generic" decorator that we wanted to be provide for others to use on any function, then we would probably try very hard to not modify the signature of the caller. But since this is only going to be used on a callable that is most likely designed to work with `bottle` and `simpl` and that `simpl` is "opinionated" in how it should be used, then I feel that adding a first argument to the callable is more **explicit** and therefore less complex and less prone to errors.

```python
@rest.body(schema=WIDGET)
def post_widget(body):
    data = save(body)
    return data
```

In following with trying to be more explicit, it might also make sense to leave a `body` argument unchanged if the call is not coming from a bottle route:
```python
def decorator(fxn):
    def wrapper(*args, **kwargs):
        # Only override body if we're in a bottle call
        if in_bottle_route_with_body():
            body = bottle.request.json
        else:
            body = args.pop(0)
        return fxn(body, *args, **kwargs)
    return wrapper
```
This way the callable can still be called directly from other code or in tests.

larsbutler: Would we do the same for the query parameters as well? What if `@rest.body` and `@rest.query` are used together? We would have to be really careful about the ordering of the decorators, since both `body` are `query` are going be JSON in 99% of cases, and it would be easy to mix them up. That's just as bad as "where did this kwarg come from?".

ziadsawalha: Good question.

Body is singular; our APIs only support one body if it is supplied. And mapping that to one arg is pretty straight-forward, intuitive, and has the benefits discussed above.

I'm not sure the same applies to query params.

To add some use cases to the discussion, here's the list of the most common query params we have in our APIs:
- pagination: `limit` & `offset` (or `marker`)
- `status` - usually a list (supports filtering on more than one) with special keys (ex. `?status=all`)
- text search (`q` param)
- `facet` key for facetted views
- `sort` key - usually multiples also and can be preceded by `-` for descending

Beyond these there are ones we don't use as frequently: `force` for a DELETE or POST, and some extra filters such as `region`, `user` or `tenantId`.

We've put those in kwargs historically, but that has resulted in many functions that look like this:
```python
def route(api_id, region=consts.DEFAULT_REGION,
          limit=consts.DEFAULT_PAGE_SIZE, offset=0,
          status=None):
    do_something()
```
That's pushing the limits of simple/readable IMO, but if you add a couple of route params (ex. id, sub_id) in there it gets too noisy. We could combine some of these into a `query` kwarg as you've done in this PR. We could also use the call context (`simpl.threadlocal`). Or we could continue what we've done above.

I'm not sure. Any suggestions?

larsbutler: In the case of query and body, I'm more in favor of passing things explicitly, rather than storing this stuff in `threadlocal`. That means that when I call this function directly, from another endpoint or from tests, I can simply pass my body and query directly.

With regards to the `limit`, `offset`, etc., I would actually be okay with collapsing all of these into a `query` kwarg, and let the schema definition handle the defaults. Then, when you want to use those parameters in the function body, you simply call `query['limit']`, and because of the schema validation, you have a guarantee that it is present and has a valid value.

larsbutler: I'm not super opinionated on whether or not we pack the `limit`, `offset`, etc. into `query`, though. I would like to hear some other opinions on this.

ziadsawalha: As I type up a trivial example below to test that it achieves:
- bottle still directly accessible
- works with other `simpl.rest` decorators 
- readable, intuitive, easy to use

It feels fine for POST:
```python
import bottle

from simpl import rest

@bottle.post('/v1/widgets/<api_id>')
@rest.query(schema={'force': bool}, defaults={'force': False})
@rest.body(schema=WIDGET)
def edit_widget(body, api_id, query=None):
    """Edit an existing widget."""
    assert query and 'force' in query, "Never, ever fails! Of course..."

    # If I want to accommodate those that say query params in POSTs are bad,
    # I can still access bottle directly.
    if 'X-Force' in bottle.request.headers:
        query['force'] = bottle.request.headers['X-Force']

    try:
        save(api_id, body, force=query['force'])
    except IndexError:
        raise rest.HTTPError("No widget with id '%s'" % api_id, http_code=404)
```

I came across issues with GET which I'll post shortly in the code.

simpl/incubator/rest.py
```diff
@@ -117,62 +117,27 @@ def validate(val):
...
-                        body = body_schema(body)
-                    except volup.MultipleInvalid as exc:
-                        raise MultiValidationError(exc.errors)
-
                 # validate the query string per the schema (if application):
                 query = bottle.request.query.dict
                 if query_schema is not None:
                     try:
-                        query = query_schema(query)
+                        query = schema(query)
```
ziadsawalha: I'm still looking for the best way to integrate with pagination and `process_params`.

Our pagination and `process_params` code is very opinionated. So keeping it out of `rest.query` would make `rest.query` more flexible, but we've been approaching `simpl` as a very opinionated library/framework, so I would say we should actually couple `rest.query` with `process_params` more tightly first and then iterate on them together. Otherwise we're introducing two different ways to handle query params, which is not opinionated.

A good first place to start would be to have `rest.query` be effectively a decorator version of `process_params` that also accepts a custom schema.

```python
@bottle.get('/v1/widgets/')
# same signature as `process_params` with the addition of a schema kwarg for
# validating any extra data
@rest.query(filter_fields=['status'], schema={'cache': str})
@rest.paginated(resource_name='widget'})
def list_widgets(status=None, limit=0, offset=0, query=None):
    """List widgets in pages filtered by optional status[es]."""
    return query_widgets(limit=limit, offset=offset, **query)
```
In the above example, all the output of `process_params` would be considered valid and only any remaining query params would be validated by the schema so that we wouldn't have to define status, limit, and offset again in query.

If `query` has all of `limit, status, offset` and `cache`, then `query` is a good name for it. But if it only has `cache` then maybe we rename it to `extra_params` which would look like this:


```python
@bottle.get('/v1/widgets/')
@rest.query(filter_fields=['status'], schema={'cache': str})
@rest.paginated(resource_name='widget'})
def list_widgets(status=None, limit=0, offset=0, extra_prams=None):
    """List widgets in pages filtered by optional status[es]."""
    assert extra_params and extra_params.keys() == ['cache']
    return query_widgets(status=status, limit=limit, offset=offset, **extra_params)
```

So, practically, I think the following would accomplish the integration:
1 - `process_params` should be extended to support a custom schema argument that is used to validate any extra params that are not part of filter_fields or standard_params.
2 - `rest.query` should have pretty much the same signature as process_params
3 - the line above should call process_params.

I think there will be opportunities to refactor while doing that and `rest.query` would become a very small wrapper for `process_params`.

larsbutler: Honestly, the only thing I see that `process_params` does for us is that it allows comma separated params: `?foo=1&foo=2,3&bar=asdf`. Why do we need this? What's wrong with `?foo=1&foo=2&foo=3&bar=asdf`?

samstav:
> we should actually couple rest.query with process_params

I actually thought through this exact thing when I saw this PR, but I thought that the purpose of the decorator and the purpose of process_params were too different. I am not so sure about this....

samstav:
> A good first place to start would be to have rest.query be effectively a decorator version of process_params that also accepts a custom schema.

This sounds fine, unless it's *only* going to be used as a decorator...

simpl/incubator/rest.py
ziadsawalha: It's easier to do as user (especially in a browser address bar) and less verbose.

ziadsawalha:
@samstav - if you'r suggesting that we keep `process_params` as a function that can be called directly (and not as a decorator), I agree. I suggested that `rest.query` will be a very light wrapper to `process_params` that makes it available as a decorator, but the original function will persist and be usable directly.

simpl/incubator/rest.py
ziadsawalha: process_params also handles our current API needs (pagination, sorting, faceting, filtering, etc...) out of the box. `rest.query` as it stands does not yet.

simpl/incubator/rest.py
samstav: It seems like some amount of this discussion is about signatures.

We expect to need to use bottle's threadlocal-based interface for accessing at least _some_ of the current request context, no matter what it attribute/value in that context we are going after. As much as I'd like for things to be explicit, this is how bottle is used.

Could we just allow the decorators to do the validation as well as mutating those globals? Then we go about our business accessing whichever pieces of the request context as usual, from these:

https://github.com/bottlepy/bottle/blob/0.12.8/bottle.py#L3691-L3698

I think this interface would be more consistent with "how to use bottle", and should relieve some of the function signature problems?

samstav:
I like the idea of being able to call `process_params` directly-- I was just asking if we would _actually_ use it that way, or we would find that we never call it directly and only use the decorator to interface with it.

ziadsawalha:
We already use it today, so keeping it available would make for an easier transition. IMO it would be good to separate the decorator logic from the bottle code.

simpl/incubator/rest.py
larsbutler: > It's easier to do as user (especially in a browser address bar) and less verbose.

It's barely less verbose, it adds special meaning to query params (which are only known by `process_params` and by nothing else), it adds unnecessary gizmos to the query interface. I would like to know if anyone else designs their query param interface this way. I've never seen anything like it. It seems like a solution to a problem that doesn't actually exist.

larsbutler: > process_params also handles our current API needs (pagination, sorting, faceting, filtering, etc...) out of the box. rest.query as it stands does not yet.

In that case, we should re-think the interface of `rest.pagination`. It was not subjected to nearly as much discussion or scrutiny as this incubation code. If we have to change things to achieve a coherent and simple design, let's change things.

Instead of reading the `limit`, `offset`, etc. params from `kwargs` or `bottle.request.query`, just read them from a `query` kwarg if present (or from bottle.request.query, in case `rest.query` is not used).

simpl/incubator/rest.py
larsbutler: As much as I am not a fan of use global variables (like bottle does), you make a very good point about signatures. So, the only parameters that an endpoint function would have would be the "tokens" defined in the route, right?

samstav: Correct, you mean this kind of thing, right? 

`/<api_id>`

larsbutler: Yeah.

larsbutler:
I've updated the implementation per @samstav's suggestion (https://github.com/checkmate/simpl/pull/69#discussion_r35232050).

larsbutler:
*bump* @samstav @ziadsawalha Is the current design of this patch acceptable?

ziadsawalha:
I still have a few concerns with this patch:

- **body**: it introduces a duplicate implementation of [body](https://github.com/checkmate/simpl/pull/69/files#diff-a79a85db87a52aac1efdd1c2899b2137R158) - we should have one. I think that might be worth putting in a separate patch to modify the existing one in a separate PR. There are other issues with the body implementation that I did not raise because we've been talking about query so much. These are things like:
    - modifying the [content of the request]() - I would prefer if the original content remains available.

larsbutler:
This PR has been stale for a while, and I haven't had time to work on it. I'm going to abandon for now.

PR #71, opened 2015-07-22
chore(build): publish universal wheel

samstav:
Run setup.py like this:

  $ python setup.py sdist bdist_wheel upload

I added some instructions to the GitHub wiki: https://github.com/checkmate/simpl/wiki#cutting-a-release

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[wheel]
+universal = 1


ziadsawalha:
You need access to https://pypi.python.org/pypi/simpl; Maintainer at least? Can we document `$ python setup.py sdist bdist_wheel upload` somewhere too along with the process for bumping the version and any tags we create?

samstav:
@ziadsawalha https://github.com/checkmate/simpl/wiki#cutting-a-release

PR #75, opened 2015-08-10
chore(version bump)

Pablosan:


Gifflen:
@Pablosan was this in response to some of the recent PRs?

Pablosan:
@Gifflen we are ready to use the new `chronos` module on one of our projects. So this is so that I can push a new version of `simpl` to Pypi.

PR #76, opened 2015-08-19
feat(mongodb): wrapper for mongodb backend

ziadsawalha:
- Ported from existing code
- tests + docs added
- manipulator logic made optional (but default)


all-requirements.txt
```diff
@@ -10,3 +10,6 @@ tornado==4.2
 
 # For simpl.middleware.cors
 WebOb==1.4.1
+
+# MongoDB
+pymongo==2.8.1
```
samstav: Preferably, we'd be able to use pymongo > 3.0. It seems like this would require some amount of work, but it might be hard to estimate how much. 

https://api.mongodb.org/python/current/changelog.html#changes-in-version-3-0

simpl/db/mongodb.py
```diff
@@ -0,0 +1,554 @@
...
+
+    def revert_key(self, key):
+        """Restore transformed key returning from database."""
+        return key.replace(self.replacement, self.replace)
+
+    def transform_incoming(self, son, collection):
+        """Recursively replace all keys that need transforming."""
+        return self._transform_incoming(copy.deepcopy(son), collection)
+
+    def _transform_incoming(self, son, collection, skip=0):
```
samstav: Confession: I contributed to this function, and I'm wary of it. On rare occasions I still see the following error come through on airbrake for our applications using this code:

```
 InvalidDocument: key 'akey.with.dots' must not contain '.'
```

simpl/db/mongodb.py
```diff
@@ -0,0 +1,554 @@
...
+        """Connect to and return mongodb database object."""
+        if self._connection is None:
+
+            self._connection = reconnector.MongoProxy(
+                self.client[self.database_name], logger=LOG)
+
+            # pymongo/database.py includes ObjectIdInjector manipulator
+            # by default. Here, we are resetting the manipulators b/c
+            # we don't want the ObjectIdInjector to add an `_id` to
+            # our modifiers.
```
samstav: After some length of time using this code, I started to wonder what we were trading off by effectively disabling the ObjectID Injector.  I don't consider myself a mongodb expert by any means, but I'd like to be fully aware of any caveats or consequences. 

My half-baked (i.e. half thought-out) suggestion would be that, if we are to leave the objectidinjector disabled by default, then we require that the id's being used are instances of pymongo's `ObjectId`. It seems like there is a lot more to it than just generating an arbitrary uuid, as we have tended to do instead. 

https://github.com/mongodb/mongo-python-driver/blob/master/bson/objectid.py#L55-L292

all-requirements.txt
ziadsawalha: Yeah. I pip installed and got 3.0.3 and that broke a ton of things. As usual, I'm trying to port existing knowledge w/o spending innovation credits :-)

simpl/db/mongodb.py
ziadsawalha: I'll dig in and try to rehydrate the reasons we added this. It may be that we're opinionated about it and we'll limit functionality or the use case was specific to one app.

simpl/db/mongodb.py
ziadsawalha: The reasoning is pretty detailed in https://github.com/checkmate/simpl/pull/76/files#diff-b304b6c35c39c7c2e29afaf7e6a295c6R116... and still valid, IMO. It's because we wanted to specify our own IDs and support partial updates.

all-requirements.txt
samstav: I am not sure I would qualify this fore **core** until it works with the latest major release of pymongo.

simpl/db/mongodb.py
samstav: I wonder if we were misusing `_id` though. It seems to me that we should have just left `_id` alone, and since we wanted our own unique ids to just use `id`. We do this anyway, and in our documents `id` ends up being the same as `_id`. 

Then, on the cases where `_id` is overwritten, we don't have to care, because our application just uses `id`, or, `uuid`, or something else.

samstav:
You do get bonus points for a `+1000 -0` Pull Request though.

all-requirements.txt
Pablosan: I think, in this particular case, we are porting outdated knowledge and therefore should not do it. Absolutely bring the learnings forward, just not the code. I'm wondering how many things we solved in the past have since been solved in pymongo?

I completely agree with @samstav: if it doesn't work with the latest pymongo, we should not include it. Further I'm guessing that, if the code were updated to be 3.0.3 compatible, it would be very difficult to identify which parts of this code are still necessary.

simpl/db/mongodb.py
```diff
@@ -0,0 +1,555 @@
...
+
+This module exposes the SimplDB class which provides an opinoinated backend
+implementation that supports using MongoDB. The opinons of the implementation
+are:
+
+- we want to just supply a URL to connect to the database.
+- we want to use your own "id" fields to key documents.
+- we want the indexes to be specified in code.
+- want to perform basic CRUD operations on collections
+- we only want to simple, JSON objects (Jdicts, lists, strings and ints)
```
larsbutler: - we only want to _use_ simple, JSON objects (Jdicts, lists, strings and ints) ?

larsbutler: Also, what is a `Jdict`?

larsbutler:
Test coverage for these new modules is somewhat low:

```
simpl.chronos              14      0   100%   
simpl.config              352     54    85%   259, 264-265, 330-333, 339, 376-378, 390, 396, 425, 429, 437, 492, 515-516, 571-575, 581-582, 598-604, 631-632, 639, 641-644, 655-656, 661, 668, 789, 813, 824-829, 848-855, 860-877, 880
simpl.db                    0      0   100%   
simpl.db.mongodb          221    103    53%   65-66, 95-109, 164, 177-179, 197-202, 213-216, 220-225, 234, 263, 285-299, 320-332, 336, 354-365, 375-384, 401, 409, 422-425, 429-432, 464, 468, 480-483, 485-486, 489, 492, 497-501, 508-509, 511, 513, 516-520, 536, 538-539, 550-555
simpl.db.reconnector       74     27    64%   37, 79-88, 92-100, 105, 109, 113, 127-128, 148, 153, 157, 161, 165
simpl.git                 299     12    96%   139, 199-200, 281, 333, 431, 541, 734-738
simpl.incubator             1      0   100%   
simpl.incubator.dicts     147      7    95%   49, 68, 147-148, 155, 248, 261
simpl.incubator.rest       59      1    98%   152
simpl.log                  49     35    29%   90-97, 107-114, 127-136, 142-156, 168-172, 177-180
simpl.rest                147      6    96%   25-26, 59, 74, 212, 355
simpl.secrets              12      4    67%   32-35
simpl.server               69     44    36%   96-122, 126-148, 152, 163-169
TOTAL                    1599    307    81%   
```

simpl/db/mongodb.py
ziadsawalha: It's a typo. Will fix.

ziadsawalha:
Yes, I've been adding tests. I just increased it from 74 to 80.

larsbutler:
> Yes, I've been adding tests. I just increased it from 74 to 80.

Cool!

simpl/db/mongodb.py
ziadsawalha: I made this optional. So the default behavior acts as is we have in prod (checkmate, opus, waldo), but you can disable all the manipulators if desired.

Again. I'm trying to port this over (with additional tests and cleanup) w/o spending "innovation credits". We can iterate/improve on it once it is here.

all-requirements.txt
ziadsawalha: I've added support for pymongo 3.0.3 (and still works with 2.8.1). The code still applies; there's no automatic reconnect, so we have an implementation for that, although I've used a more modern one that addresses issues we had with cursors.. Much of the text search ad cursor code enables pagination and works with rest.py. That's all still valid.

samstav: Excellent, @ziadsawalha.

PR #77, opened 2015-08-20
fix(rest): 416 on unexpected content-types

larsbutler:
Because this explicit error isn't raised, I found subtle bugs in an
application that uses this. If the client forgot to specify the
content-type in the request header, the server would process the request
as if everything was fine, except that it would ignore the content. The
was leading to incorrect behavior in the server application.

### DIFF
--- a/simpl/rest.py
+++ b/simpl/rest.py
@@ -62,6 +62,12 @@ def body(schema=None, types=None, required=False, default=None):
         """Return a decorated callable."""
         def wrapped(*args, **kwargs):
             """Callable to called when the decorated function is called."""
+            if bottle.request.content_type not in types:
+                content_type_msg = (
+                    "Unexpected Content-Type '%s'. Supported Content-Types: %s"
+                    % (bottle.request.content_type, types)
+                )
+                bottle.abort(416, content_type_msg)
             data = bottle.request.json
             if required and not data:
                 bottle.abort(400, "Call body cannot be empty")
--- a/tests/test_rest.py
+++ b/tests/test_rest.py
@@ -19,6 +19,7 @@ import unittest
 import bottle
 import mock
 import six
+import voluptuous as volup
 
 from simpl import rest
 import webtest
@@ -28,8 +29,11 @@ class TestBodyDecorator(unittest.TestCase):
 
     """Tests for :func:`simpl.rest.body`."""
 
-    def test_decoration(self):
+    @mock.patch.object(rest.bottle, 'request')
+    def test_decoration(self, mock_request):
         """Test decorated function is called."""
+        mock_request.json = None
+        mock_request.content_type = 'application/json'
         mock_handler = mock.Mock(return_value='X')
         decorated = rest.body()(mock_handler)
         self.assertTrue(callable(decorated))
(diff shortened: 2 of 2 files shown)


samstav:
> incorrect behavior in the server application

What was the incorrect behavior?

simpl/rest.py
```diff
@@ -62,6 +62,12 @@ def wrap(fxn):
         """Return a decorated callable."""
         def wrapped(*args, **kwargs):
             """Callable to called when the decorated function is called."""
+            if bottle.request.content_type not in types:
+                content_type_msg = (
+                    "Unexpected Content-Type '%s'. Supported Content-Types: %s"
+                    % (bottle.request.content_type, types)
+                )
+                bottle.abort(416, content_type_msg)
```
samstav: This will 416 fail if the `Content-Type` header was not supplied?

larsbutler:
> What was the incorrect behavior?

Incorrect was the following:
- API endpoint uses `rest.body` and includes a schema for the request body.
- The request body schema has an optional integer parameter.
- Default contents were also defined, including a default value for the integer param above.
- If the client passes in a JSON blob in the request body, but fails to include the `content-type: application/json` (I was the client, on the command, using cURL--I forgot the content-type), the request will be accepted anyway.
- Because it is accepted, `rest.body` gets the `bottle.request.json`, which is `None` if the content-type != application/json.
- Thus, the request.body was set to the default values.
- Thus, even though I had specified an explicit value, it was silently ignored.

samstav:
I see, thanks @larsbutler.

Relevant code in bottle:

https://github.com/bottlepy/bottle/blob/0.12.8/bottle.py#L1112-L1124

ziadsawalha:
_Should_ we support a POST without Content-Type? We could do that by not using bottle's `.json`...

samstav:
In another project, I added some middleware (perhaps questionable) that went something like this:

```python
if 'content-type' not in request.headers:
    try:
        json.loads(request.body)
    except StandardError:
        return
    else:
        request.headers['Content-Type'] = 'application/json'
```

This seems like an ok thing to do, but only if the loaded json could be reused downstream.

ziadsawalha:
It's not mandatory. Per the HTTP 1.1 specification:

> Any HTTP/1.1 message containing an entity-body SHOULD include a Content-Type header field defining the media type of that body. If and only if the media type is not given by a Content-Type field, the recipient MAY attempt to guess the media type via inspection of its content and/or the name extension(s) of the URI used to identify the resource. If the media type remains unknown, the recipient SHOULD treat it as type "application/octet-stream".

We're allowed to guess it if we want...

larsbutler:
> Should we support a POST without Content-Type? We could do that by not using bottle's .json...

I don't think so; if we're already going to be so strict and explicit about our APIs, request content schema, etc., it would only make sense to me to check and enforce the content-type.

If a client sends us JSON, but they fail to tell us that it's `application/json`, or if the content-type they give us is something we don't expect, we should say, "Yo, your request invalid."

ziadsawalha:
Given the HTTP specs says:
> the recipient MAY attempt to guess the media type via inspection of its content

It would certainly be the better user experience to try to guess it. And it would not be difficult on the backend.

samstav:
> If a client sends us JSON, but they fail to tell us that it's `application/json`

I agree with @ziadsawalha that we should try to guess this

> or if the content-type they give us is something we don't expect, we should say, "Yo, your request invalid."

I agree with you here.

PR #78, opened 2015-08-25
Add Text Search Parsing

ziadsawalha:
When a text index exists on a collection and we want to search it using a `q=` query param, these patches can do that.

### DIFF
--- a/simpl/db/mongodb.py
+++ b/simpl/db/mongodb.py
@@ -115,6 +115,55 @@ def scrub(data):
         raise ValidationError("Input '%s' not permitted: %s" % (data, exc))
 
 
+def build_text_search(strings, name_field='name'):
+    """Build mongodb query that performs text search for string(s).
+
+    This is the backend implementation of the front-end search box on a list.
+    It performs text search on a text index and a regex search on a `name`
+    field. The value(s) for this are parsed by `rest.process_params` from the
+    `q` query param.
+
+    :param list strings: strings to search on.
+    :keyword str name: field to search on in addition to search index.
+
+    For example, searching the following collection for 'john':
+
+    #  Name    Description (in text index)
+    -  ------  ---------------------------
+    1  John    Big boss
+    2  Johnny  Boss' son
+    3  Thomas  John's first employee
+    4  Henry   He quit
+
+    Search returns all records except the last one:
+    #1 - match on name
+    #2 - partial match on name
+    #3 - text index match on description
+
+    >>> import pprint
+    >>> pprint.pprint(build_text_search(['john']))
+    {'$or': [{'$text': {'$search': 'john'}},
+             {'name': {'$options': 'i', '$regex': 'john'}}]}
+
+    >>> import pprint
+    >>> pprint.pprint(build_text_search(['john', 'tom'], name_field='objName'))
+    {'$or': [{'$text': {'$search': 'john tom'}},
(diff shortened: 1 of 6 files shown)


test-requirements.txt
```diff
@@ -7,6 +7,7 @@ mock
 mongobox
 nose
 nose-ignore-docstring
+prettyprint
```
samstav: Dev crumb?

ziadsawalha: No. It's needed for the doctests.

simpl/db/mongodb.py
```diff
@@ -115,6 +115,55 @@ def scrub(data):
         raise ValidationError("Input '%s' not permitted: %s" % (data, exc))
 
 
+def build_text_search(strings, name_field='name'):
```
samstav: Was just looking over the `$text` documentation, seems like it would be a nice feature and pretty easy to add negations to your string, specified by another keyword argument on this function?

http://docs.mongodb.org/master/reference/operator/query/text/#exclude-documents-that-contain-a-term

samstav: And what about text score? This seems useful. Maybe `sort=True` could return the results sorted highest-to-lowest by textscore?

http://docs.mongodb.org/master/reference/operator/query/text/#text-score

We could revisit this later too, but I assume these will be requests from the UI and should probably be sorted by a score like this.

samstav: Just to be clear, I am basically just brainstorming here, not suggesting that these necessarily be implemented before we merge this.

samstav:
Your metasyntactic variables are jolly. +1

simpl/db/mongodb.py
ziadsawalha: These are all valid use cases, but would go beyond what we have in prod today. So I would request we (you :smile:) add them as issues (feature requests) and not add to this PR.

Note that the current implementation is not mongo-specific on the front end (just one or more strings coming in as `?q=search1,search2&q=search3`). We talked about implementing a github-like query syntax (ex. `no:milestone`). That might conflict with mongo-style negations if we implement them now.

ziadsawalha:
Thank you :-)

samstav:
Cool. LGTM. I will give others a chance to review.

/cc @Pablosan @larsbutler @Gifflen @wanghaven @ryandub @griffinb

simpl/db/mongodb.py
```diff
@@ -115,6 +115,55 @@ def scrub(data):
...
+    {'$or': [{'$text': {'$search': 'john'}},
+             {'name': {'$options': 'i', '$regex': 'john'}}]}
+
+    >>> import pprint
+    >>> pprint.pprint(build_text_search(['john', 'tom'], name_field='objName'))
+    {'$or': [{'$text': {'$search': 'john tom'}},
+             {'$or': [{'objName': {'$options': 'i', '$regex': 'john'}},
+                      {'objName': {'$options': 'i', '$regex': 'tom'}}]}]}
+    """
+    assert isinstance(strings, list)
```
griffinb: What happens if `strings` isn't a list?

griffinb:
+1 looks good.

simpl/db/mongodb.py
ziadsawalha: It throws the error and fails.

PR #79, opened 2015-08-25
Config Fixes: `strict` parsing and deprecation warning

ziadsawalha:


simpl/config.py
```diff
@@ -656,6 +672,14 @@ def parse_passthru_args(argv):
             return argv[:dashdash], argv[dashdash + 1:]
         return argv, []
 
+    def set_keyring_value(self, key, value, namespace=None):
```
samstav: Cool. Can you leave an example on how/why to use this?

ziadsawalha: Oops - this was meant to go i a separate PR.

simpl/config.py
```diff
@@ -259,10 +259,7 @@ def add_argument(self, parser, permissive=False, **override_kwargs):
                 kwargs['help'] = "%s (or set %s)" % (kwargs['help'],
                                                      kwargs['env'])
             if permissive:
-                try:
-                    required = kwargs.pop('required', None)
-                except KeyError:
-                    pass
+                required = kwargs.pop('required', None)
```
samstav: Good catch- that the try/except is not needed.

samstav:
LGTM, you finished with this one?

ziadsawalha:
Yes. Ready.

PR #80, opened 2015-08-27
feat(middleware): add context middleware

ziadsawalha:


simpl/middleware/context.py
```diff
@@ -0,0 +1,90 @@
...
+                url += environ['HTTP_HOST']
+            else:
+                url += environ['SERVER_NAME']
+
+                if environ['wsgi.url_scheme'] == 'https':
+                    if environ['SERVER_PORT'] != '443':
+                        url += ':' + environ['SERVER_PORT']
+                else:
+                    if environ['SERVER_PORT'] != '80':
+                        url += ':' + environ['SERVER_PORT']
```
samstav: Just an FYI on how bottle does this, it might be more robust?

https://github.com/bottlepy/bottle/blob/master/bottle.py#L1329-L1340

Then accesses the rebuilt url with `geturl()`

https://github.com/bottlepy/bottle/blob/master/bottle.py#L1321

ziadsawalha: Thanks, @samstav, I've used the same code. Given we don't need the path, I didn't think we needed the join/geturl part. How does that look?

samstav: Cool, do you think it's an improvement? LGTM

ziadsawalha: It's more robust (I took out to tests for failure that are now covered by defaults). I trust that bottle has been around the block. So Yes.

README.md
```diff
@@ -44,6 +44,7 @@ Includes sample middleware for use with WSGI apps including bottle.
 
 Middleware included:
 - CORS: handles CORS requests
+- Context: handles setting a threadlocal context with with transaction ID.
```
samstav: What about:

> handles setting a threadlocal call context and sets the Transaction ID

ziadsawalha: Edited - thanks!

samstav:
Cool. I will give others a chance to review.

ziadsawalha:
Rebased off of master.

PR #82, opened 2015-09-01
simpl server

samstav:
This change installs a `simpl` command into your environment during package installation using the `console_scripts` entry point. The first available subcommand is `server`. You can now use `simpl server` to quickly run your app.

The interface is similar to that of [bottle's "command line interface"](http://bottlepy.org/docs/dev/tutorial.html#command-line-interface).

#### 1. Point `simpl server` to your routes
Example:

```python
# test_app.py

import bottle

@bottle.get('/')
def hello():
    return 'Hello World!'
```

Run the command

```
simpl server -a test_app -p 8888
Bottle v0.12.8 server starting up (using XTornadoServer())...
Listening on http://127.0.0.1:8888/
Hit Ctrl-C to quit.
```

test it...

```
you@localhost $ http get localhost:8888
HTTP/1.1 200 OK
Content-Length: 12
Content-Type: text/html; charset=UTF-8
Server: TornadoServer/4.2

Hello World!
```

Of course, this might not be how you choose to continue as your app matures

#### 2. Attach simpl's `server` subcommand to your parser

Admittedly, this technique is a little strange, and the code must be in the exact order as written below. But it works.

```python
import argparse

from simpl import server


parser = argparse.ArgumentParser()
server_parser = server.attach_parser(parser.add_subparsers())
server_parser.set_defaults(func=server.run)
args = awesome_parser.parse_args()

conf = config.init(options=server.OPTIONS)
conf.parse()
args._func(conf)
```

#### 3. Just use the server.OPTIONS, and run bottle however you want.

```python
import bottle
from simpl import server

conf = server.CONFIG
conf.parse()

b = bottle.Bottle()
# add middleware, routes to Bottle instance
conf['app'] = b
server.run(conf)
```

--------

### --help

```
you@localhost$ simpl server --help
usage: simpl server [-h] [--ini PATH] [--app APP] [--host HOST] [--port PORT]
                    [--server SERVER] [--debug-server] [--quiet-server]
                    [--no-reloader] [--interval INTERVAL]
                    [--adapter-options [ADAPTER_OPTIONS [ADAPTER_OPTIONS ...]]]

optional arguments:
  -h, --help            show this help message and exit

initialization (metaconfig) arguments:
  evaluated first and can be used to source an entire config

  --ini PATH            Source some or all of the options from this ini file.

Server Options:
  --app APP, -a APP     WSGI application to load by name.
                        Ex: package.module gets the module
                            package.module:name gets the variable 'name'
                            package.module.func() calls func() and gets the result
  --host HOST           Server address to bind to. (default: 127.0.0.1)
  --port PORT, -p PORT  Server port to bind to. (default: 8080)
  --server SERVER, -s SERVER
                        Server adapter to use. To see more, run:
                        `python -c "import bottle;print(bottle.server_names.keys())"`
                         (default: xtornado)
  --debug-server        Run bottle server with debug=True which is useful
                        for development or troubleshooting. Warning: This
                        may expose raw tracebacks and unmodified error
                        messages in responses! Note: this is not an option
                        to configure DEBUG level logging. (default: False)
  --quiet-server        Suppress bottle's output to stdout and stderr,
                        e.g. "Bottle v0.12.8 server starting up..." and
                        others. (default: False)
  --no-reloader         Disable bottle auto-reloading server, which
                        automatically restarts the server when file
                        changes are detected. Note: some server adapters,
                        such as eventlet, do not support auto-reloading. (default: False)
  --interval INTERVAL, -i INTERVAL
                        Auto-reloader interval in seconds (default: 1)
  --adapter-options [ADAPTER_OPTIONS [ADAPTER_OPTIONS ...]], -o [ADAPTER_OPTIONS [ADAPTER_OPTIONS ...]]
                        Key-value pairs separated by '=' to be passed to
                        the underlying server adapter, e.g. XEventletServer,
                        and are mapped to the adapter's self.options
                        instance attribute. Example usage:
                          simpl server -s xeventlet -o keyfile=~/mykey ciphers=GOST94
```

### DIFF
--- a/simpl/server.py
+++ b/simpl/server.py
@@ -254,3 +254,52 @@ class XTornadoServer(bottle.ServerAdapter):  # pylint: disable=R0903
         tornado.ioloop.IOLoop.instance().start()
 
 bottle.server_names['xtornado'] = XTornadoServer
+
+
+def attach_parser(subparser):
+    """Given a subparser, build and return the server parser."""
+    return subparser.add_parser(
+        'server',
+        help='Run a bottle based server',
+        parents=[
+            CONFIG.build_parser(
+                add_help=False,
+                # might need conflict_handler
+            ),
+        ],
+    )
+
+
+def run(conf):
+    """Simpl server command line interface."""
+    # TODO(sam): Need a way to pass arbitrary kwargs
+    # to server handlers.
+
+    # bottle_kwargs = {key.replace('-', '_'): val
+    #                  for key, val in zip(extras[::2], extras[1::2])}
+
+    try:
+        if conf.reloader and not os.getenv('BOTTLE_CHILD'):
+            LOG.info("Running bottle server with reloader.")
+        return bottle.run(
+            app=conf.app,
+            server=conf.server,
+            host=conf.host,
+            port=conf.port,
+            interval=conf.interval,
+            reloader=conf.reloader,
+            quiet=conf.quiet,
+            debug=conf.debug,
+        )
+    except KeyboardInterrupt:
+        sys.exit("\nKilled simpl server.")
+
+
+def main(argv=None):
+    """Entry point for server, runs based on parsed CONFIG."""
+    CONFIG.parse(argv=argv)
+    return run(CONFIG)
+
+
+if __name__ == '__main__':
+    main()
--- a/simpl/server.py
+++ b/simpl/server.py
@@ -269,8 +269,54 @@ def attach_parser(subparser):
     )
 
 
+def fmt_pairs(obj, indent=4, sort_key=None):
+    """Format and sort a list of pairs, usually for printing.
+
+    If sort_key is provided, the value will be passed as the
+    'key' keyword argument of the sorted() function when
+    sorting the items. This allows for the input such as
+    [('A', 3), ('B', 5), ('Z', 1)] to be sorted by the ints
+    but formatted like so:
+
+        l = [('A', 3), ('B', 5), ('Z', 1)]
+        print(fmt_pairs(l, sort_key=lambda x: x[1]))
+
+            Z 1
+            A 3
+            B 5
+        where the default behavior would be:
+
+        print(fmt_pairs(l))
+
+            A 3
+            B 5
+            Z 1
+    """
+    lengths = [len(x[0]) for x in obj]
+    if not lengths:
+        return ''
+    longest = max(lengths)
+    obj = sorted(obj, key=sort_key)
+    formatter = '%s{: <%d} {}' % (' '*indent, longest)
+    string = '\n'.join([formatter.format(k, v) for k, v in obj])
+    return string
+
+
+def fmt_routes(bottle_app):
+    """Return a pretty formatted string of the list of routes."""
+    routes = [(r.method, r.rule) for r in bottle_app.routes]
+    if not routes:
+        return
+    string = 'Routes:\n'
+    string += fmt_pairs(routes, sort_key=operator.itemgetter(1))
+    return string
+
+
 def run(conf):
-    """Simpl server command line interface."""
+    """Run server based on this configuration.
+
+    Expects configuration options defined in server.OPTIONS
+    """
     if isinstance(conf.adapter_options, list):
         options = {key: val for _dict in conf.adapter_options
                    for key, val in _dict.items()}
@@ -283,9 +329,22 @@ def run(conf):
     if conf.app and (os.getcwd() not in sys.path):
         sys.path.append(os.getcwd())
 
+    # get wsgi app the same way bottle does *in advance*
+    conf.app = conf.app or bottle.default_app()
+    if isinstance(conf.app, six.string_types):
+        conf.app = bottle.load_app(conf.app)
+
     try:
-        if conf.reloader and not os.getenv('BOTTLE_CHILD'):
-            LOG.info("Running bottle server with reloader.")
+        if os.getenv('BOTTLE_CHILD'):
+            if conf.reloader:
+                LOG.info("Running bottle server with reloader.")
+            if conf.app and not conf.quiet:
+                try:
+                    routes = fmt_routes(conf.app)
+                    if routes:
+                        print('\n{}'.format(fmt_routes(conf.app)), end='\n\n')
+                except AttributeError:
+                    pass
         return bottle.run(
             app=conf.app,
             server=conf.server,
@@ -302,7 +361,7 @@ def run(conf):
 
 
 def main(argv=None):
-    """Entry point for server, runs based on parsed CONFIG."""
+    """Command line entry point for server, runs based on parsed CONFIG."""
     CONFIG.parse(argv=argv)
     return run(CONFIG)
 
--- a/simpl/server.py
+++ b/simpl/server.py
@@ -59,7 +59,7 @@ OPTIONS = [
         group='Server Options',
     ),
     config.Option(
-        '--debug', '-d',
+        '--debug-server',
         help=_fill(
             'Run bottle server with debug=True which is useful for '
             'development or troubleshooting. Warning: This may expose raw '
@@ -70,7 +70,7 @@ OPTIONS = [
         group='Server Options',
     ),
     config.Option(
-        '--quiet', '-q',
+        '--quiet-server',
         help=_fill(
             'Suppress bottle\'s output to stdout and stderr, e.g. '
             '"Bottle v0.12.8 server starting up..." and others.'),
@@ -288,8 +288,8 @@ def run(conf):
             port=conf.port,
             interval=conf.interval,
             reloader=conf.reloader,
-            quiet=conf.quiet,
-            debug=conf.debug,
+            quiet=conf.quiet_server,
+            debug=conf.debug_server,
         )
     except KeyboardInterrupt:
         sys.exit("\nKilled simpl server.")
--- a/simpl/server.py
+++ b/simpl/server.py
@@ -292,6 +292,11 @@ def run(conf):
         options = {}
     else:
         options = copy.copy(conf.adapter_options)
+
+    # waiting for https://github.com/bottlepy/bottle/pull/783
+    if conf.app and (os.getcwd() not in sys.path):
+        sys.path.append(os.getcwd())
+
     try:
         if conf.reloader and not os.getenv('BOTTLE_CHILD'):
             LOG.info("Running bottle server with reloader.")
--- a/simpl/server.py
+++ b/simpl/server.py
@@ -39,13 +39,13 @@ OPTIONS = [
     ),
     config.Option(
         '--host',
-        help='Proxy server address to bind to.',
+        help='Server address to bind to.',
         default='127.0.0.1',
         group='Server Options',
     ),
     config.Option(
         '--port', '-p',
-        help='Proxy server port to bind to.',
+        help='Server port to bind to.',
         type=int,
         default=8080,
         group='Server Options',
--- a/simpl/server.py
+++ b/simpl/server.py
@@ -14,6 +14,7 @@
 
 """Bottle Server Module."""
 
+import copy
 import logging
 import os
 import sys
@@ -86,7 +87,7 @@ OPTIONS = [
         help=_fill(
             'Disable bottle auto-reloading server, which automatically '
             'restarts the server when file changes are detected. Note: '
-            'some server handlers, such as eventlet, do not support '
+            'some server adapters, such as eventlet, do not support '
             'auto-reloading.'),
         group='Server Options',
     ),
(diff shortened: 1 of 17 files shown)


ryandub:
+1 from me!

ziadsawalha:
Nice! A couple of notes:
- could you add docs to the README?
- Is it possible call it from a project's entry point instead of using the simpl commandline and entry point? I.e. right now you show how to use the simpl command-line separately and pass it any entry point (could be anything, not even in the project that is using simpl). But for many of our projects we use simpl.config and have many options that we control from inside the project (like customizing the [program name](https://github.com/checkmate/simpl/pull/82/files#diff-f2e2993939c76e748c03c4f993dc8d58R24)) and we have project-specific entry points. For those projects, we might want to append a set of server options and then expose the start command from the project itself. For example:

```python
"""Project Awesome."""

from simpl import config
from simpl import log
from simpl import server

SIMPL_OPTIONS = log.OPTIONS + server.OPTIONS
OPTIONS = MY_OPTIONS + SIMPL_OPTIONS

if __name__ == "__main__":
    conf = config.init(options=OPTIONS)
    log.configure(conf)
    conf.parse()
    if conf.command == "server":
        server.run(name="awesome", conf)
    else:
        do_something_else()
```

So the command would be similar to what you have in the PR description, but would be the project entry point, no simpl's:
```
$ awesome server --help
usage: awesome server [-h] [--ini PATH] [--app APP] [--host HOST] [--port PORT]
                    [--server SERVER] [--debug-server] [--quiet-server]
                    [--no-reloader] [--interval INTERVAL]
                    [--adapter-options [ADAPTER_OPTIONS [ADAPTER_OPTIONS ...]]]

optional arguments:
  -h, --help            show this help message and exit

initialization (metaconfig) arguments:
  evaluated first and can be used to source an entire config

  --ini PATH            Source some or all of the options from this ini file.

Server Options:
  --app APP, -a APP     WSGI application to load by name.
                        Ex: package.module gets the module
                            package.module:name gets the variable 'name'
                            package.module.func() calls func() and gets the result
  --host HOST           Server address to bind to. (default: 127.0.0.1)
  --port PORT, -p PORT  Server port to bind to. (default: 8080)
  --server SERVER, -s SERVER
                        Server adapter to use. To see more, run:
                        `python -c "import bottle;print(bottle.server_names.keys())"`
                         (default: xtornado)
  --debug-server        Run bottle server with debug=True which is useful
                        for development or troubleshooting. Warning: This
                        may expose raw tracebacks and unmodified error
                        messages in responses! Note: this is not an option
                        to configure DEBUG level logging. (default: False)
  --quiet-server        Suppress bottle's output to stdout and stderr,
                        e.g. "Bottle v0.12.8 server starting up..." and
                        others. (default: False)
  --no-reloader         Disable bottle auto-reloading server, which
                        automatically restarts the server when file
                        changes are detected. Note: some server adapters,
                        such as eventlet, do not support auto-reloading. (default: False)
  --interval INTERVAL, -i INTERVAL
                        Auto-reloader interval in seconds (default: 1)
  --adapter-options [ADAPTER_OPTIONS [ADAPTER_OPTIONS ...]], -o [ADAPTER_OPTIONS [ADAPTER_OPTIONS ...]]
                        Key-value pairs separated by '=' to be passed to
                        the underlying server adapter, e.g. XEventletServer,
                        and are mapped to the adapter's self.options
                        instance attribute. Example usage:
                          simpl server -s xeventlet -o keyfile=~/mykey ciphers=GOST94
```

If that is possible, could you show an example of you recommend it be done? If not, can we explore making that possible?

samstav:
@ziadsawalha Thanks for the comments.

Thinking/coding out loud...

```python
"""Project Awesome."""

from simpl import config
from simpl import log
from simpl import server
from simpl.utils import cli as cli_utils

SIMPL_OPTIONS = log.OPTIONS + server.OPTIONS
OPTIONS = MY_OPTIONS + SIMPL_OPTIONS
# Where MY_OPTIONS have some server options in
# with group='Server Options'

# Also, project awesome defines the `awesome`
# console_scripts entry_point which points to
# this file's main()

def main():

    conf = config.init(options=OPTIONS)
    log.configure(conf)
    conf.parse()

    awesome_parser = cli_utils.HelpfulParser(prog='awesome')
    server_parser = server.attach_parser(awesome_parser.add_subparsers())
    server_parser.set_defaults(_func=server.main)

    args = awesome_parser.parse_args()
    args._func(conf)


if __name__ == "__main__":

    main()
```

Maybe we could improve on this?

ziadsawalha:
> Where MY_OPTIONS have some server options with group='Server Options

Is that needed if we provide `server.OPTIONS` in simpl? See `SIMPL_OPTIONS = log.OPTIONS + server.OPTIONS`

samstav:
@ziadsawalha it's not needed, but it would allow your apps options to be grouped with the server options provided by simpl in the help/usage output.

simpl/server.py
```diff
@@ -14,13 +14,111 @@
...
+    config.Option(
+        '--server', '-s',
+        help=('Server adapter to use. To see more, run:\n'
+              '`python -c "import bottle;print('
+              'bottle.server_names.keys())"`\n'),
+        default='xtornado',
+        group='Server Options',
+    ),
+    config.Option(
+        '--debug-server',
```
ziadsawalha: I noticed that you changed this from `--debug\-d` to `--debug-server`. Why not just have it enabled when `--debug\-d` is supplied? I'm not sure I see the need/benefit for `--debug\-d` and `--debug-server`  being separate.

`--debug\-d` already generates gobs of data (as might be expected from a debug mode), so having it output traces in bottle also seems acceptable. _Note: for the record, --verbose is how to log heavily but not dump gobs of data out._

I do recognize the challenge of addressing the duplicate `--debug\-d` option in server.OPTIONS and log.OPTIONS if we did that. IMO, we should try to solve that and keep only one `-d` flag.

simpl/server.py
```diff
@@ -14,13 +14,111 @@
...
+            'Run bottle server with debug=True which is useful for '
+            'development or troubleshooting. Warning: This may expose raw '
+            'tracebacks and unmodified error messages in responses! Note: '
+            'this is not an option to configure DEBUG level logging.'),
+        default=False,
+        action='store_true',
+        group='Server Options',
+    ),
+    config.Option(
+        '--quiet-server',
```
ziadsawalha: Same with `--debug\-d`, this should just piggyback off of `--quiet\-q`.

samstav:
@ziadsawalha I am not sure how to handle the conflicting options. I believe `--debug` and `--debug-server` should mean different things, so I am not sure about merging them. And if I did merge them, where would the option live? In `server.py` or `log.py` ? What would the `description` be? Or in a common options module?

I think `--debug` WRT the bottle server should (read: _will eventually_) mean that the traceback will be returned in the json payload if `FormatExceptionMiddleware` catches it.

ziadsawalha:
To resolve the conflict, `--debug\-d` could move into config.py and become a built-in option in config.py. If putting simpl in debug mode starts to mean more than just logging, then it would no longer be appropriate for that option to be in log.py.

> I believe --debug and --debug-server should mean different things

So would we have `--debug-db`, `--debug-config`, etc... for each module?

ziadsawalha:
Thanks for documenting "Attach simpl's server subcommand to your parser". Could you add that markdown to the code in the PR as well? Either in the server module docstrings or a readme? It should end up being the canonical "how to use this" text.

tests/test_server.py
```diff
@@ -44,6 +47,16 @@ def test_xeventlet(self):
         self.assertTrue(resp.ok)
         self.assertEqual(resp.content, b'<b>Hello xeventlet</b>!')
 
+    def test_simpl_server(self):
+        argv = ['server', '--quiet-server', '--port', str(get_free_port())]
+        proc = multiprocessing.Process(
+            target=simpl_cli.main, kwargs={'argv': argv})
```
ziadsawalha: I struggled with this in https://github.com/checkmate/simpl/blob/master/tests/test_server.py#L109. I think this might not get correctly reflected in in test coverage.

tests/test_server.py
```diff
@@ -44,6 +47,16 @@ def test_xeventlet(self):
...
+        proc.start()
+        time.sleep(.5)
```
ziadsawalha: Is 0.5 seconds a guess on how long it will take the process to start? Maybe combining https://github.com/samstav/simpl/blob/simpl-server/tests/test_server.py#L112 with a timeout would be a more reliable approach? The test should not fail if the process takes longer than 0.5s and it should not wait idly for 0.5 seconds if the process is up in 10ms.

samstav: Good point, I'll fix this up.

samstav:
> So would we have --debug-db, --debug-config, etc... for each module?

Hmm, yeah, this doesn't sound great. I like the built-in option idea above, so I'll change it to that.

samstav:
Either in the server module docstrings or a readme?

I like the idea of the examples in the README. I'll put 'em there

simpl/server.py
```diff
@@ -169,3 +267,59 @@ def run(self, handler):
...
+            server=conf.server,
+            host=conf.host,
+            port=conf.port,
+            interval=conf.interval,
+            reloader=conf.reloader,
+            quiet=conf.quiet_server,
+            debug=conf.debug_server,
+            **options
+        )
+    except KeyboardInterrupt:
```
samstav: bottle already catches this.

ziadsawalha: I think the bottle file watcher (for reloading) issues KeyboardInterrupt to stop the process that is running the old set of files. Myabe you need to check if you're running in the server or watcher process (see https://github.com/checkmate/simpl/pull/82/files#diff-7f2f24942936abfb2a1b215c243bd28bR301).

simpl/server.py
```diff
@@ -169,3 +267,59 @@ def run(self, handler):
...
+        options = {}
+    else:
+        options = copy.copy(conf.adapter_options)
+
+    # waiting for https://github.com/bottlepy/bottle/pull/783
+    if conf.app and (os.getcwd() not in sys.path):
+        sys.path.append(os.getcwd())
+
+    try:
+        if conf.reloader and not os.getenv('BOTTLE_CHILD'):
```
samstav: I understand why I used `not os.getenv('BOTTLE_CHILD')`, but I should probably remove the `not`.

simpl/server.py
```diff
@@ -169,3 +267,59 @@ def run(self, handler):
...
+            CONFIG.build_parser(
+                add_help=False,
+                # might need conflict_handler
+            ),
+        ],
+    )
+
+
+def run(conf):
+    """Simpl server command line interface."""
```
samstav: This docstring is now a bit inaccurate

simpl/server.py
```diff
@@ -169,3 +267,59 @@ def run(self, handler):
...
+            quiet=conf.quiet_server,
+            debug=conf.debug_server,
+            **options
+        )
+    except KeyboardInterrupt:
+        sys.exit("\nKilled simpl server.")
+
+
+def main(argv=None):
+    """Entry point for server, runs based on parsed CONFIG."""
```
samstav: This is more like "the command line interface"

https://github.com/checkmate/simpl/pull/82/files#r39530041

simpl/utils/cli.py
```diff
@@ -0,0 +1,65 @@
...
+                0, os.path.basename(target) or os.path.relpath(target))
+            message = ("%s. Try getting help with `%s --help`"
+                       % (message, " ".join(sys.argv)))
+        if print_help:
+            self.print_help()
+        sys.stderr.write('\nerror: %s\n' % message)
+        sys.exit(2)
+
+
+def kwarg(string, separator='='):
```
samstav: It might be convenient for this function to accept a `type` keyword argument, which would cast all values thru this type on their way out.

samstav:
TODO:

Add pretty-formatted-print of routes on startup when in debug mode:

```python

def fmt_pairs(obj, indent=4, sort_key=None):
    """Format and sort a list of pairs, usually for printing.
    If sort_key is provided, the value will be passed as the
    'key' keyword argument of the sorted() function when
    sorting the items. This allows for the input such as
    [('A', 3), ('B', 5), ('Z', 1)] to be sorted by the ints
    but formatted like so:
        l = [('A', 3), ('B', 5), ('Z', 1)]
        print(fmt_pairs(l, sort_key=lambda x: x[1]))
            Z 1
            A 3
            B 5
        where the default behavior would be:
        print(fmt_pairs(l))
            A 3
            B 5
            Z 1
    """
    longest = max(six.moves.map(len, [x[0] for x in obj]))
    obj = sorted(obj, key=sort_key)
    formatter = '%s{: <%d} {}' % (' '*indent, longest)
    string = '\n'.join([formatter.format(k, v) for k, v in obj])
    return string


def fmt_routes(bottle_app):
    """Return a pretty formatted string of the list of routes."""
    routes = [(r.method, r.rule) for r in bottle_app.routes]
    string = 'Routes:\n'
    string += fmt_pairs(routes, sort_key=operator.itemgetter(1))
    return string

# then, once the bottle app is built:
print('\n{}'.format(fmt_routes(bottle_app)), end='\n\n')

# you get something like this

Routes:
    GET    /
    POST  /api/resources/<resource_id>
    GET    /api/resources/<resource_id>
    OPTIONS /gui
```

PR #84, opened 2015-09-07
fix(context): clear call context when done

ziadsawalha:
Some threads can be reused by a bottle server and
old call data can persist in the thread-local
storage. This ensures the data is cleared after
each request.

### DIFF
--- a/simpl/middleware/context.py
+++ b/simpl/middleware/context.py
@@ -32,6 +32,7 @@ Example usage:
     bottle.run(app=chain)
 """
 
+import contextlib
 import logging
 import uuid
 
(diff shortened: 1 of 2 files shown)


samstav:
The threadlocal call context will now be unavailable to any middlewares "above" the `ContextMiddleware`, right? Whereas previously, the context was available for the life of the thread. I don't think we've ever had code in that logical place that would depend on the threadlocal call context, but I wanted to make sure I understood correctly and document the caveat if there is one.

ziadsawalha:
The threadlocal data is available to anything in the thread. However, the .default() context will be cleared after the ContextMiddleware finishes handling its response. I make the assumption that any workers or asynchronous code running is in another thread/process/machine and would have had the context copied over by then.

Given the ContextMiddleware is what returns the X-Transaction-Id header, it should be very close to the top of the stack (if not at the top). Since any error or failure should go back with the X-Transaction-Id header in it as an ID we can use for further troubleshooting.

ziadsawalha:
Thanks @samstav

PR #89, opened 2015-10-01
fix(config): fix order of precedence in option processing

larsbutler:
Default values were overriding values set at higher levels of
precedence, namely command line args (which have the highest
precedence!).

Patch includes specific tests cases, which could be expanded to include
ini and keyring sources of config options. (The tests only cover
defaults, env, and command line sources of config params.)

### DIFF
--- a/simpl/config.py
+++ b/simpl/config.py
@@ -477,8 +477,15 @@ class Config(collections.MutableMapping):
         options = []
         for option in self._options:
             kwargs = option.kwargs.copy()
-            if kwargs.get('default') is None:
+            original_default = kwargs.get('default')
+            if original_default is not None:
                 kwargs['default'] = argparse.SUPPRESS
+                # Since we're suppressing defaults, we need to somehow preserve
+                # the default value for the help text:
+                help_text = kwargs.get('help', '')
+                kwargs['help'] = '%s (default: %s)' % (
+                    help_text, original_default
+                )
             temp = Option(*option.args, **kwargs)
             options.append(temp)
         parser = self.build_parser(options, permissive=permissive)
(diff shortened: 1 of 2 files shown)


simpl/config.py
```diff
@@ -477,8 +477,15 @@ def parse_cli(self, argv=None, permissive=False):
...
-            if kwargs.get('default') is None:
+            original_default = kwargs.get('default')
+            if original_default is not None:
                 kwargs['default'] = argparse.SUPPRESS
+                # Since we're suppressing defaults, we need to somehow preserve
+                # the default value for the help text:
+                help_text = kwargs.get('help', '')
+                kwargs['help'] = '%s (default: %s)' % (
+                    help_text, original_default
+                )
```
samstav: This changed from

```python
if kwargs.get('default') is None:
    kwargs['default'] = argparse.SUPPRESS
```

to

```python
if kwargs.get('default') is not None:
    kwargs['default'] = argparse.SUPPRESS
    # then add default and help text manually
```

What's going on here? First change I can think of is that default values of `None` will be included in the help/usage output now.

samstav:
Nice catch @larsbutler!

Is there a test case for

> For example, if the defaults
> define a value for the option `foo`, and `args` contains `foo=None`,
> the `args` should override the default.

?

And I left a comment demonstrating my confusion about `argparse.SUPPRESS` and the help/usage output. Maybe you could include one test that asserts that the `--help` output includes/doesn't-include what you expect?

tests/test_config.py
```diff
@@ -430,5 +430,63 @@ def test_default_help_formatter(self, mock_stdout):
 """
         self.assertEqual(mock_stdout.getvalue(), expected)
 
+
+class TestConfigPrecedence(unittest.TestCase):
```
ziadsawalha: We've been missing these tests. Glad to see them come in...

simpl/config.py
```diff
@@ -477,8 +477,15 @@ def parse_cli(self, argv=None, permissive=False):
         options = []
         for option in self._options:
             kwargs = option.kwargs.copy()
-            if kwargs.get('default') is None:
```
ziadsawalha: `argparse.SUPPRESS is not None` - so the deleted line handled `default=argparse.SUPPRESS` correctly, I think and the new one may miss it?

PR #91, opened 2015-10-12
fix(rest): filters must come back in mongo format

ziadsawalha:
Since we’re coding this specifically to work with
bottle and mongodb, multiple filters should be
formatted as a mongodb $in filter condition.


samstav:
Does this change imply that any comma separated query parameters coming through an API call for a simpl-based app must be going to a mongo query? If so, is that ok? At this point, nothing in `rest.py` is this strictly coupled with mongo.

ziadsawalha:
We've only worked with (and assumed) mongodb as the backend. But, to keep `simpl.rest` useful without mongodb, I've moved the mongo-specific code to the mongo module.

PR #95, opened 2015-10-15
Config: Refactor of CLI Parsing, Validation, and Precedence Handling

ziadsawalha:
This refactor/enhancement to `simpl.config` should help address recurring issues and increasing complexity in that module caused, IMO, but trying to overload parsing and validation in `parse_cli`. It's been problematic - see #89, #79 among other bugs.

### DIFF
--- a/simpl/config.py
+++ b/simpl/config.py
@@ -405,7 +405,7 @@ class Config(collections.MutableMapping):
         self._parser_kwargs.setdefault('parents', [])
         self._parser_kwargs['parents'].append(metaparser)
         self._metaconf._values = self._metaconf.load_options(
-            argv=argv, permissive=True)
+            argv=argv)
         self._metaconf.provision(self)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


simpl/config.py
```diff
@@ -466,12 +467,107 @@ def build_parser(self, options, permissive=False, **override_kwargs):
...
+
     def parse_cli(self, argv=None, permissive=False):
         """Parse command-line arguments into values.
 
         :keyword permissive: when true, does not validate required or extra
             arguments.
         """
+        warnings.warn("Calling `cli_values` directly is deprecated. Call "
+                      "`cli_values` and/or `validate_config` instead.",
+                      DeprecationWarning)
```
samstav: I think you meant `` Calling `parse_cli` directly is deprecated. `` Maybe also mention it is scheduled for removal in 1.0.0 or something?

simpl/config.py
```diff
@@ -466,12 +467,107 @@ def build_parser(self, options, permissive=False, **override_kwargs):
...
+        Parses arguments provided on the command-line (or sys.argv). Only
+        returns arguments that are explicitly supplied, so we strip out
+        defaults and validation rules like `required` in this call.
+        """
+        options = []
+        for option in self._options:
+            kwargs = option.kwargs.copy()
+            # Must explicitly set default to None or `store_true` and
+            # `store_false` actions will set the value to true or false,
+            # respectively.
```
samstav: Interesting..........

samstav:
Would differentiating between `None` and "the value was not supplied" be helpful for simpl/config?

For example

```python
NOTSET = object()

argparse.ArgumentParser(argument_default=NOTSET)
```

simpl/config.py
ziadsawalha: Yes! I'll fix.

ziadsawalha:
WRT https://github.com/checkmate/simpl/pull/95#issuecomment-148410650

I've been thinking about that since you mentioned it. I'm trying to think of a use case where None would be insufficient. None, IMO, is the pythonic way of saying NOTSET. What do we gain by having that extra concept and code? How would we use it?

ziadsawalha:
Thanks, @samstav

PR #96, opened 2015-10-19
FormatExceptionMiddleware

### LINKED ISSUE #83: Proposed FormatExceptionMiddleware
How does this look?

Simpl's rest.HTTPError would change to look like this:

```python
class SimplHTTPError(SimplException):

    default_status = 500

    def __init__(self, status=None, body=None, exception=None,
                 traceback=None, **options):
        body = body if body is not None else ''
        super(SimplHTTPError, self).__init__(body)
        self.exception = exception
        self.traceback = traceback
        self.body = body
        self.status_code = status or self.default_status
        self.options = options
```

And then middleware:

```python
import json
import traceback as tb_mod

import bottle


class FormatExceptionMiddleware(object):

    """Format outgoing exceptions.

    Uses and is compatible-with bottle exception formatting.

    - Handle Bottle Exceptions (even when catchall=False).
    - Handle SimplHTTPError
    - Fail-safe to a generic error (unexpected_error)

    This code in this middleware is meant to be generic.
    Don't catch and translate fancy exceptions here, do
    it in the application logic or a separate middleware and
    raise a bottle.HTTPError or a SimplHTTPError.
    """

    unexpected_error = "We're sorry, something went wrong."

    def __init__(self, app, conf):
        """Initialize class."""
        self.app = app
        self.config = conf
        LOG.debug("Added middleware: %s", type(self).__name__)

    def __call__(self, environ, start_response):
        """Catch exceptions and format them based on config."""
        try:
            response = self.app(environ, start_response)
            return response
        except bottle.HTTPError as error:
            LOG.error("Formatting a bottle exception.",
                      exc_info=error)

            self._format_error_response(error)
            start_response(error.status_line, error.headerlist)
            return error
        except SimplHTTPError as error:
            LOG.error("Formatting a SimplHTTPError exception.",
                      exc_info=error)
            error = bottle.HTTPError(
                status=error.status_code, body=error.body,
                exception=error.exception or error,
                traceback=error.traceback or tb_mod.format_exc())
            self._format_error_response(error)
            start_response(error.status_line, error.headerlist)
            return error
        except Exception as error:  # pylint: disable=W0703
            LOG.error("Formatting an unexpected exception.",
                      exc_info=error)
            error = bottle.HTTPError(
                status=500, body=self.unexpected_error,
                exception=error,
                traceback=tb_mod.format_exc())
            self._format_error_response(error)
            start_response(error.status_line, error.headerlist)
            return error

    @classmethod
    def format_error_response(cls, error, debug=False):
        """Format error responses properly."""
        status_code = error.status_code or 500
        data = {
            'code': status_code,
            'description': error.body or cls.unexpected_error,
            'message': bottle.HTTP_CODES.get(status_code, u''),
        }
        if bottle.DEBUG and debug:
            LOG.warning("Debug-mode server is returning traceback and error "
                        "details in the response with a %s status.",
                        error.status_code)
            if error.exception:
                data['exception'] = repr(error.exception)
            else:
                data['exception'] = repr(error)
            data['traceback'] = error.traceback or tb_mod.format_exc()

        # overwrite previous body attr with json
        if isinstance(data['description'], bytes):
            data['description'] = data['description'].decode(
                'utf-8', errors='replace')

        error.body = [json.dumps(data, indent=4).encode('utf8')]
        error.set_header('Content-Type', 'application/json')
        return error.body

    def _format_error_response(self, error):
        """Modify the error response object to include a json body."""
        return type(self).format_error_response(
            error, debug=self.config.debug)
```

And then we can drop the [error formatter function](https://github.com/checkmate/simpl/blob/v0.7.1/simpl/rest.py#L306-L368) in rest.py and instead just do this when we setup the server:

```python
import bottle

app = bottle.default_app()
app.catchall = False
# format_error_response is now a classmethod
app.default_error_handler = FormatExceptionMiddleware.format_error_response
app = FormatExceptionMiddleware(app)
```

This gets us the following:

* consolidated logic
* catch and format *all* errors with consistent logic (bottle's catchall can be True or False, your preference)
* deals well with errors that occur in other middleware!

Still need to [address the fact that this assumes json](https://github.com/checkmate/simpl/issues/83#issuecomment-138233259).

samstav:
Includes an updated error formatting function and custom httperror class.

We need to deprecate/remove and maybe add backwards compatibility for the following:

  - https://github.com/checkmate/simpl/blob/v0.7.1/simpl/rest.py#L306-L368
  - https://github.com/checkmate/simpl/blob/v0.7.1/simpl/rest.py#L33-L41

The current pull request is an update/upgrade/refactor of these.

The suggested use of bottle for consistent error handling and formatting becomes:

```python
app = bottle.default_app()
app.catchall = False
app.default_error_handler = rest.format_error_response
app = FormatExceptionMiddleware(app)
bottle.run(app=app)
```

This allows your app to handle unexpected exceptions in your routehandler logic OR in other middleware. Turning catchall off is beneficial in case any error handling logic needs access to the exception context, i.e. the return value of `sys.exc_info()`. If bottle always handles your errors, `sys.exc_info()` always returns `(None, None, None)`.

Fixes #83

### DIFF
--- a/simpl/exceptions.py
+++ b/simpl/exceptions.py
@@ -121,3 +121,28 @@ class SimplConfigUnknownOption(SimplConfigException):
     For example, a specified ini file has an option with no corresponding
     config.Option.
     """
+
+
+class SimplHTTPError(SimplException):
+
+    """A custom http error class inspired by bottle's HTTPError.
+
+    Bottle has special treatment of its own HTTPErrors. Instances of
+    this class will be ignored by bottle. The benefit is that the
+    error handling logic in rest.py can deal with either exception
+    using the same code.
+
+    See rest.format_error_response()
+    """
+
+    default_status = 500
+
+    def __init__(self, status=None, body=None, exception=None,
+                 traceback=None, **options):
+        body = body if body is not None else ''
+        super(SimplHTTPError, self).__init__(body)
+        self.exception = exception
+        self.traceback = traceback
+        self.body = body
+        self.status_code = int(status) or self.default_status
+        self.options = options
--- a/simpl/exceptions.py
+++ b/simpl/exceptions.py
@@ -132,7 +132,7 @@ class SimplHTTPError(SimplException):
     error handling logic in rest.py can deal with either exception
     using the same code.
 
-    See rest.format_error_response()
+    See rest.httperror_handler
     """
 
     default_status = 500
--- a/simpl/middleware/errors.py
+++ b/simpl/middleware/errors.py
@@ -1,0 +1,173 @@
+# Copyright (c) 2011-2015 Rackspace US, Inc.
+# All Rights Reserved.
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+"""Middleware for errors and exceptions.
+
+Example:
+
+    # Allow this middleware to catch, handle and format all errors.
+
+    import bottle
+    from simpl.middleware import errors
+
+    app = bottle.default_app()
+    app.catchall = False
+    # To catch errors that occur in other middleware...
+    app = FormatExceptionMiddleware(app)
+    # Tell bottle to use rest.format_error_response when
+    # handling its own bottle.HTTPErrors
+    app.default_error_handler = rest.format_error_response
+    bottle.run(app=app)
+
+
+    Example responses (the following use catchall=False)
+    -----------------------------------------------------
+
+    An unexpected error if the server is in debug mode:
+
+        # raise ValueError()
+        {
+            "code": 500,
+            "description": "Internal Server Error",
+            "exception": "ValueError()",
+            "message": "Internal Server Error",
+            "traceback": "Traceback (most recent call last): ... ",
+        }
+
+
+    An "expected" error if the server is in debug mode:
+
+        # raise SimplHTTPError(body='Hey!', status=405)
+        {
+            "code": 405,
+            "description": "Hey!",
+            "exception": "SimplHTTPError('Hey!',)",
+            "message": "Method Not Allowed",
+            "traceback": "Traceback (most recent call last): ... ",
+        }
+
+
+    Unexpected error without debug mode:
+
+        # raise ValueError()
+        {
+            "code": 500,
+            "description": "We're sorry, something went wrong.",
+            "message": "Internal Server Error"
+        }
+
+
+    An "expected" error without debug mode:
+
+        # raise SimplHTTPError(body='Hey!', status=405)
+        {
+            "code": 405,
+            "description": "Hey!",
+            "message": "Method Not Allowed"
+        }
+"""
+
+import logging
+import traceback as tb_mod
+
+import bottle
+
+from simpl import exceptions as simpl_exc
+from simpl import rest
+
+LOG = logging.getLogger(__name__)
+
+
+def _catchall_enabled(app):
+    """Check the bottle app for catchall."""
+    while hasattr(app, 'app'):
+        if isinstance(app, bottle.Bottle):
+            break
+        app = app.app
+    if hasattr(app, 'catchall'):
+        return app.catchall
+    else:
+        return bottle.default_app().catchall
+
+
+class FormatExceptionMiddleware(object):
+
+    """Format outgoing exceptions.
+
+    Uses and is compatible-with bottle exception formatting.
+
+    - Handle Bottle Exceptions (even when catchall=False).
+    - Handle SimplHTTPError
+    - Fail-safe to a generic error (unexpected_error)
+
+    The code in this middleware is meant to be generic.
+    Don't catch and translate fancy exceptions here, do
+    it in the application logic or a separate middleware and
+    raise a bottle.HTTPError or a SimplHTTPError.
+    """
+
+    def __init__(self, app, conf=None):
+        """Initialize class."""
+        if _catchall_enabled(app) is True:
+            LOG.warning("Bottle's catchall flag is enabled for your app. "
+                        "The suggested use for this middleware is with "
+                        "catchall disabled, otherwise bottle will overwrite "
+                        "the exception context, sys.exc_info() and raise a "
+                        "bottle.HTTPError instead, which renders the error "
+                        "handling middleware less useful and less flexible.")
+        self.app = app
+        self.config = conf
+        LOG.debug("Added middleware: %s", type(self).__name__)
+
+    def __call__(self, environ, start_response):
+        """Catch exceptions and format them based on config."""
+        try:
+            return self.app(environ, start_response)
+        except bottle.HTTPError as error:
+            LOG.error("Formatting a bottle exception.",
+                      exc_info=error)
+
+            rest.format_error_response(error)
+            start_response(error.status_line, error.headerlist)
+            return error
+        except simpl_exc.SimplHTTPError as error:
+            LOG.error("Formatting a SimplHTTPError exception.",
+                      exc_info=error)
+            error = bottle.HTTPError(
+                status=error.status_code, body=error.body,
+                exception=error.exception or error,
+                traceback=error.traceback or tb_mod.format_exc())
+            rest.format_error_response(error)
+            start_response(error.status_line, error.headerlist)
+            return error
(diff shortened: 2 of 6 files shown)


simpl/middleware/errors.py
```diff
@@ -0,0 +1,144 @@
...
+    def __init__(self, app, conf=None):
+        """Initialize class."""
+        self.app = app
+        self.config = conf
+        LOG.debug("Added middleware: %s", type(self).__name__)
+
+    def __call__(self, environ, start_response):
+        """Catch exceptions and format them based on config."""
+        try:
+            LOG.warning("Ohai from middleware.")
```
ziadsawalha: Are you sure you want this to be a warning?

simpl/middleware/errors.py
```diff
@@ -0,0 +1,144 @@
...
+            LOG.error("Formatting a SimplHTTPError exception.",
+                      exc_info=error)
+            error = bottle.HTTPError(
+                status=error.status_code, body=error.body,
+                exception=error.exception or error,
+                traceback=error.traceback or tb_mod.format_exc())
+            rest.format_error_response(error)
+            start_response(error.status_line, error.headerlist)
+            return error
+        except Exception as error:  # pylint: disable=W0703
```
ziadsawalha: How will this impact SystemExit or KeyboardInterrupts?

What are the chances of an Exception being raised that is not inheriting from Exception? That may be bad practice, but should we also catch it using, maybe, an `except:`?

simpl/middleware/errors.py
samstav: OOps, leftover from debugging!

simpl/middleware/errors.py
samstav: That is a good question. Funny that you ask, because I was just experimenting with this today. I assume if we catch one of these errors we want to re-raise it after we send the response? I think I'm off track..

samstav: https://github.com/checkmate/simpl/commit/34d5a6968bcfd75d967314f7ad47f51c3695e007

ziadsawalha: I think if we have on of these two exceptions (SystemExit and KeyboardInterrupt) then we can assume the system is being shut down. Unless we are going to get into a complicated solution, it is probably safer to not return from the call and just reraise so that the system can continue to shut down.

```
except (SystemExit, KeyboardInterrupt):
    LOG.error("System shutting down in the middle of handling a call.")
    raise
```

samstav: I agree that we can assume the system is being shut down. Although... bottle doesn't?

Run this:

```python
from __future__ import print_function
import bottle, sys

app = bottle.default_app()

def keyboard():
    raise KeyboardInterrupt()
app.route(path='/keyboard', callback=keyboard)

def sysexit():
    sys.exit()
app.route(path='/sysexit', callback=sysexit)

print('\n'.join([j.method + '  ' + j.rule for j in APP.routes]))
bottle.run(app, debug=True)
```

If you call those routes, `curl localhost:8080/sysexit` or `curl localhost:8080/keyboard` the server stays up and running and returns you the `text/plain` content `A server error occurred.  Please contact the administrator.` 

If the bottle server is going to stay up and running when we re-raise _anyway_, we might as well return an error response in our canonical format?

ziadsawalha: That's a good point.

If I were going to argue for "the more correct" handling or better interop between our code and bottle or other libs, I'd want to raise those exceptions and make sure Bottle handled them as one would expect. Particularly if there is an attempt to shut the process down. But I agree that returning a consistently formatted error trumps any haggling over that.

simpl/exceptions.py
```diff
@@ -121,3 +121,28 @@ class SimplConfigUnknownOption(SimplConfigException):
     For example, a specified ini file has an option with no corresponding
     config.Option.
     """
+
+
+class SimplHTTPError(SimplException):
```
ziadsawalha: Can we combine this with https://github.com/checkmate/simpl/blob/master/simpl/rest.py#L33? Having two will cause confusion as to which to use.

simpl/middleware/errors.py
```diff
@@ -0,0 +1,192 @@
...
+                traceback=tb_mod.format_exc())
+            rest.format_error_response(error)
+            start_response(error.status_line, error.headerlist)
+            return error
+
+
+if __name__ == '__main__':
+    # pylint: disable=missing-docstring
+    logging.basicConfig()
+    APP = bottle.Bottle(catchall=False)
```
ziadsawalha: I think it's fine to put a test or example usage here, but this code seems specific to the discussion about SystemExit and KeyboardInterrupt. I'm not sure they remain relevant to stay in the code base.

simpl/middleware/errors.py
```diff
@@ -0,0 +1,192 @@
...
+            "code": 405,
+            "message": "Hey!",
+            "reason": "Method Not Allowed"
+        }
+"""
+from __future__ import print_function
+
+import logging
+import sys
+import traceback as tb_mod
```
ziadsawalha: I don't see a name conflict that would require the rename here. Could this stay as traceback? `import traceback`

samstav: +1

simpl/middleware/errors.py
samstav: +1

simpl/middleware/errors.py
```diff
@@ -0,0 +1,192 @@
...
+
+import bottle
+
+from simpl import exceptions as simpl_exc
```
ziadsawalha: Same here. Although `exceptions` is more generic, there is only one exceptions module in simpl. `from simpl import exceptions`?

samstav: +1

simpl/middleware/errors.py
```diff
@@ -0,0 +1,192 @@
...
+from simpl import rest
+
+LOG = logging.getLogger(__name__)
+
+
+def _catchall_enabled(app):
+    """Check the bottle app for catchall."""
+    while hasattr(app, 'app'):
```
ziadsawalha: This assume every middleware in the chain has an `app` attribute that points to the downstream call handler. I think that's a convention we use (and maybe others). Would explicitly passing the bottle app be more explicit?

```python
next_app = FormatExceptionMiddleware(next_app, bottle.default_app(), conf=conf)
```

simpl/middleware/errors.py
```diff
@@ -0,0 +1,192 @@
...
+    it in the application logic or a separate middleware and
+    raise a bottle.HTTPError or a SimplHTTPError.
+    """
+
+    def __init__(self, app, conf=None):
+        """Initialize class."""
+        if _catchall_enabled(app) is True:
+            LOG.warning("Bottle's catchall flag is enabled for your app. "
+                        "The suggested use for this middleware is with "
+                        "catchall disabled.")
```
ziadsawalha: What happens if it is not set? Could that be added to the message?

tests/test_middleware_errors.py
```diff
@@ -0,0 +1,325 @@
...
+        self.assertEqual(
+            resp.json_body['reason'],
+            "Internal Server Error"
+        )
+        self.assertEqual(
+            resp.json_body['code'],
+            500
+        )
+
+
```
ziadsawalha: extra line break here

tests/test_middleware_errors.py
```diff
@@ -0,0 +1,325 @@
...
+            resp.json_body['reason'],
+            "Forbidden"
+        )
+        self.assertEqual(
+            resp.json_body['code'],
+            403
+        )
+
+
+class WorstMiddlewareEver(object):
```
ziadsawalha: Hah

simpl/middleware/errors.py
samstav: That would be more explicit, but I don't think it would be worth the trouble or modifying the signature of the middleware for.

ziadsawalha: How about checking bottle.default_app() as a failsafe if following th chain doesn't get you to a bottle instance?

samstav: Sounds good to me!

simpl/middleware/errors.py
samstav: For posterity:

```python

if __name__ == '__main__':
    # pylint: disable=missing-docstring
    logging.basicConfig()
    APP = bottle.Bottle(catchall=False)

    def _keyboard():
        raise KeyboardInterrupt()
    APP.route(path='/keyboard', callback=_keyboard)

    def _sysexit():
        sys.exit()
    APP.route(path='/sysexit', callback=_sysexit)

    def _typeerror():
        raise TypeError('Poor type choice.')
    APP.route(path='/typeerror', callback=_typeerror)

    def _simplhttperror():
        raise exceptions.SimplHTTPError(body='Hey!', status=405)
    APP.route(path='/simplerror', callback=_simplhttperror)

    print('\n'.join([j.method + '  ' + j.rule for j in APP.routes]))
    APP = FormatExceptionMiddleware(APP)
    bottle.run(APP, debug=True)
```

Issue #88, opened 2015-09-30
Except block in incubator.rest catches all exceptions


ryandub:
https://github.com/checkmate/simpl/blob/master/simpl/incubator/rest.py#L189-L190

This generic exception swallows any exceptions I raise in my wrapped route:

```
import myexchandler as exc

@bottle.post('/myroute')
@rest.schema(body_schema=MYSCHEMA, body_required=True,
                        default_body={})
def some_route(body, query):
  raise exc.HTTPError(body="I'M A TEAPOT!", status=418)
```

I would expect this to raise a 418 error. However, it will always raise a 400 because the error raised here is caught by the `except` in the wrapper and then a `bottle.abort(400...)` is called.

Is this `try`/`except` block needed? Can it be taken out or narrowed down?

ryandub:
Meant to ping @larsbutler as I think you probably have the most perspective on what was going on there.

ryandub:
@larsbutler @samstav @ziadsawalha  - I think there is a use case for the incubator validation code in some of the work @Pablosan is doing, but this issue could stand in the way. Any thoughts on a solution? In order to get this working the way I would expect it to, I just removed the whole try/except block (https://github.com/ryandub/simpl/commit/5bce9d8bacdc2146854f828e28a62c74e94ecf85), but I think that has its own implications.

Is there an ideal solution here?

samstav:
@larsbutler @ryandub @ziadsawalha 

I think we should just change the

```python
except Exception as exc:
```

to

```python
except MultiValidationError as exc:
```

ryandub:
@samstav What if it isn't valid JSON? Should https://github.com/checkmate/simpl/blob/master/simpl/incubator/rest.py#L160 be wrapped as well?

samstav:
@ryandub good question.... Would that raise a ValueError ?

Pablosan:
@samstav @ryandub I think we could solve this in the `incubator` code by following the approach here: https://github.com/checkmate/simpl/blob/master/simpl/rest.py#L39

Specifically, this moves the catch-all so that it only catches on a single line (the line that validates against the schema)

However, I am pretty sure that a call to `bottle.request.json` can result in an exception, so we should probably address that in both locations (`rest.body` and `incubator.rest.schema`).

[...]

PR #101, opened 2015-11-18
fix(rest): return 400 when body gets bad content

ziadsawalha:
We were getting 500’s instead.


samstav:
@ziadsawalha did you check out this issue? https://github.com/checkmate/simpl/issues/88

ziadsawalha:
@samstav - good point. This PR tackles the json piece as suggested in the discussion on #88, am I right? If so, it can be merged separately?

I can tackle #88 in a follow up PR (today, hopefully). Works?

samstav:
@ziadsawalha Ok cool. :hammer:

ziadsawalha:
Thanks!

PR #108, opened 2016-09-08
fix(rest): Properly wrap rest.body decorator

szilveszter:
Without using `functools.wraps`, the decorator will mask the underlying stack of functions.


larsbutler:
LGTM

